PR #17, opened 2019-06-18
Fix bug with CSSStyleSheet.insertRule in IE

jiggum:
In IE, CSSStyleSheet.insertRule must has `index` argument
https://developer.mozilla.org/ko/docs/Web/API/CSSStyleSheet/insertRule

Without this param, IndexSizeError will be executed
IndexSizeError

### DIFF
--- a/src/Style.tsx
+++ b/src/Style.tsx
@@ -13,7 +13,7 @@ export const VirtuosoStyle: FC<{
       position: sticky;
       position: -webkit-sticky;
       z-index: 2;
-    } `)
+    } `, 0)
 
     style.current = styleEl
 


petyosi:
Thanks for that! I haven't tested how the component works in IE - does this fix make it work? 

P.S. I will merge this next week when having access to Windows and release a new version.

jiggum:
For this moment, this project not support IE without Babel.
However, even if after compile this library with Babel, It's not work with causes from those code.
**This fix and using Babel makes it work fine on IE11, 10**.
I'll plan to use this 'virtual list' in our project after fix this bug.
(Your codes are awesome! I was impressed about have used AATree and observable pattern)

petyosi:
Hey, @jiggum - I just published v0.7.1 which includes your fix. Also, I upgraded the `tsdx` dev dependency to the latest version, which means that you no longer need to transpile the module (at least from what I tested). Let me know if this is not the case, though. Cheers!

PR #22, opened 2019-07-26
feat: allow custom item keys via `computeItemKey`

johanneslumpe:
This PR addresses an issue regarding sorting lists. Using raw indexes as keys is discouraged in react and giving consumers the ability to provide custom keys for each row allows for stable reconciliation. Without it issues like incorrectly updated DOM nodes will pop up and cause broken UI.

@petyosi Would be awesome to get this merged. I added this functionality to our local fork but figured it would make sense for everyone who has the use case of lists which are being sorted.

### DIFF
--- a/src/Virtuoso.tsx
+++ b/src/Virtuoso.tsx
@@ -15,6 +15,7 @@ export interface VirtuosoProps {
   topItems?: number
   footer?: () => ReactElement
   item: (index: number) => ReactElement
+  computeItemKey: (index: number) => number
   itemHeight?: number
   endReached?: (index: number) => void
   scrollingStateChange?: (isScrolling: boolean) => void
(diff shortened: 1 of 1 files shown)


petyosi:
Thank you @johanneslumpe! This is a valid case. I have merged the proposed change. I will document the feature and publish a new version along with a few other changes. 

Regards,

johanneslumpe:
Awesome, thank you!

leroydev:
`computeItemKey` is non optional, which makes this PR a breaking change because you must pass `computeItemKey` or else it won't build:
```ts
computeItemKey: (index: number) => number
```

Was this intentional? I can make a PR to fix it otherwise. :)

petyosi:
Should be optional! Fix PR is welcome.

johanneslumpe:
Sorry, my bad!

leroydev:
@johanneslumpe no problem! 😃

petyosi:
v0.10.2 is published. Thank you @leroydev !

PR #23, opened 2019-07-30
fix: "ResizeObserver loop limit exceeded"

FezVrasta:
Copies the approach used in https://github.com/hshoff/vx/pull/335 to fix a "ResizeObserver loop limit exceeded" error

### DIFF
--- a/src/Utils.tsx
+++ b/src/Utils.tsx
@@ -1,4 +1,4 @@
-import { useRef, useState, useLayoutEffect } from 'react'
+import { useRef, useState, useEffect, useLayoutEffect } from 'react'
 import ResizeObserver from 'resize-observer-polyfill'
 import { TInput, TOutput } from './rxio'
 
@@ -14,12 +14,13 @@ type UseHeight = (
 export const useHeight: UseHeight = (input, onMount, onResize) => {
   const ref = useRef<CallbackRefParam>(null)
   const currentHeight = useRef(0)
+  const animationFrameID = useRef<number>(0)
   const observer = new ResizeObserver(entries => {
     const newHeight = entries[0].contentRect.height
     if (currentHeight.current !== newHeight) {
       currentHeight.current = newHeight
       if (onResize) {
-        onResize(entries[0].target as HTMLElement)
+        animationFrameID.current = window.requestAnimationFrame(() => onResize(entries[0].target as HTMLElement))
       }
       input(newHeight)
     }
@@ -37,6 +38,8 @@ export const useHeight: UseHeight = (input, onMount, onResize) => {
       ref.current = null
     }
   }
+
+  useEffect(() => () => window.cancelAnimationFrame(animationFrameID.current), [])
   return callbackRef
 }
 


petyosi:
Hey @FezVrasta - do you have a repro for the issue somewhere? Thanks!

FezVrasta:
No I'm sorry, it happened in a fairly complex non-OSS app, but you should be able to track down the issue following the PR I linked in the commit

petyosi:
Hey @FezVrasta , 

I tested the change locally and it broke the [grouped example](https://virtuoso.dev/grouped-with-load-on-demand/). I traced the problem to some height measurements being dropped due to the component being unmounted and canceling the animation frame request.

What worked for me was to remove the 
```js  
useEffect(() => () => window.cancelAnimationFrame(animationFrameID.current))
```
call. I *think* this should bring no harm, but I might be wrong (for example, measuring items which are no longer in the DOM).  We will see.

Can you please test this in your setup? Thank you.

FezVrasta:
I can't seem to be able to reproduce the problem if I comment out that line, but I'm not sure what could happen if `onResize` gets called while the component is unmounted. If that callback performs a check to make sure it's not running on a detached/destroyed node then everything should be okay.

petyosi:
Yeah, It is reproducible only in grouped mode with footer set, since multiple measurements are performed. I will examine it further and figure out something.

FezVrasta:
How do you run the docs locally with a local version of react-virtuoso? Do I have to switch to `gh-pages` and manually copy the build?

petyosi:
This is the site: https://github.com/petyosi/react-virtuoso-site

I keep it as a peer directory of the react-virtuoso repo, (the dir is called lib):
https://github.com/petyosi/react-virtuoso-site/blob/master/package.json#L32
To test the lib changes with the demos, I run `yarn start` in the `lib` directory,  and then `yarn start` in the site directory. 

Thank you for your effort!

FezVrasta:
Thanks I got it working, but I can't reproduce any issue with the grouped example, what's the issue I should see?

petyosi:
On my side, when I load the example (without any interaction - resize, scroll, etc), the items do not get loaded. I only see the header and the button. The reason is that one measurement gets dropped.

FezVrasta:
May I know more about your environment? I wonder if it's OS/browser/version specific.

petyosi:
Chrome 75 on MacOS.

FezVrasta:
Ok, same here, I'm not sure what's going on then...

petyosi:
Did you rebase on v0.10? I made one minor change in the hooks behavior there, and that is the only problem I can think of. 

Anyway, will check again on my side later this week.

FezVrasta:
I just rebased but I can't still reproduce your issue

petyosi:
Whatever it was, it works now. Sorry for the hassle. published v0.10.1 with the fix. 

Thank you once again,

FezVrasta:
I'm experiencing the same issue again, I guess I didn't fix it completely 😢

petyosi:
I haven't changed anything in the area since this PR. Something which might help me is the stack trace (and, of course, a repro would be awesome). Contact me on my email if I can observe the problem somewhere.

Cheers.

Issue #25, opened 2019-07-31
Expose callback prop for `itemsRendered`


petyosi:

joshacheson:
@petyosi are you open to having someone else implement this? I could be up for giving it a shot late this week or next week.

petyosi:
@joshacheson That would be awesome! Give it a shot.

joshacheson:
Heya! I'm working on this today (And on other days if necessary :)). I'm curious how you make use of `playground/` for testing? I'm not sure what browser environment allows you to directly import / use `.tsx`.

petyosi:
Hey @joshacheson , 

you just run `parcel playground/index.html` and you are good to go.

PR #28, opened 2019-08-12
Add `itemsRendered` prop to `<Virtuoso />` and `<GroupedVirtuoso />`

joshacheson:
This PR is for https://github.com/petyosi/react-virtuoso/issues/25.

At this moment I'd consider this just a proposal/clarification of needs, and I want to explain of the existing decisions. I'm leaving comments/questions inline.


src/Virtuoso.tsx
```diff
@@ -49,10 +53,22 @@ export const VirtuosoPresentation: FC<TVirtuosoPresentationProps> = ({
...
+    let dispose: TSubscription
+    if (itemsInView) {
+      dispose = contextValue.inView$.subscribe(itemsInView)
+    }
+    return () => {
+      if (dispose) {
+        dispose()
+      }
+    }
+  }, [itemsInView])
```
joshacheson: This effect for subscribing/unsubscribing was placed in `VirtuosoPresentation` because this meant logic did not need to be duplicated across `Virtuoso` and `GroupedVirtuoso`. I think it's... sort of odd that there's a component called `...Presentation` that now has logic for a side effect in it, but doing it this way saved us from code duplication. Curious what you think about this.

src/VirtuosoStore.tsx
```diff
@@ -168,7 +168,10 @@ const VirtuosoStore = ({ overscan = 0, totalCount = 0, itemHeight }: TVirtuosoCo
     minListIndex$,
     totalCount$,
     offsetList$
-  ).pipe(scan(listScanner(overscan), []))
+  )
+
+  const inView$ = baseList$.pipe(scan(listScanner(0), []))
+  const list$ = baseList$.pipe(scan(listScanner(overscan), []))
```
joshacheson: 1) I called this `inView$` because in my own case what I'm looking for is not which items are rendered, but which are actually in view. This needed to ignore `overscan`. This might not be the best way to factor code to achieve this result.
2) I chose the name `inView` instead of `itemsRendered` to indicate that this list tells you which items you should be seeing, but in fact which items are rendered might be a much larger list.

src/VirtuosoStore.tsx
```diff
@@ -228,6 +231,8 @@ const VirtuosoStore = ({ overscan = 0, totalCount = 0, itemHeight }: TVirtuosoCo
   const isScrolling$ = buildIsScrolling(scrollTop$)
 
   return {
+    inView$,
```
joshacheson: I exposed this stream directly rather than using `makeOutput` so that any thing that has a reference to this can handle its own subscription/disposing of subscription. When I looked at `makeOutput` it seemed as though unsubscribing was closed over, and would only occur when the function returned by `makeOutput` was called a second, third, fourth, etc. time. I might be misunderstanding how `makeOutput` works, though, and could of course refactor this to use `makeOutput`.

playground/index.tsx
```diff
@@ -43,6 +43,9 @@ const App = () => {
           style={{ height: '400px', width: '350px' }}
           overscan={300}
           groupCounts={groupCounts}
+          itemsInView={items => {
+            console.log(items)
+          }}
```
joshacheson: I will remove this change if/when we get towards merging this PR.

joshacheson: (or I could leave this in and find some way to render which items are in view)

petyosi:
Thank you very much for the request and for the comments left :). It helped a lot!

- You correctly spotted something which I have been ignoring for some time; disposal of subscriptions to the store streams when the component is unmounted. Right now, unsubscribing works by passing `undefined` to the output. This approach (passing undefined, or some variation of it, like returning the unsubscribe handle) can be used to dispose all outputs when the component is unmounted - a thing that should be done if more outputs are exposed as props. 

With that in min, (and this goes beyond the scope of this PR), I would rather have the stream being exposed through the `makeOutput` wrapper. This will allow us to handle its disposition easily just like the rest of the streams. Moving to `makeOutput` will also eliminate the `useEffect` bit in the Presentation component, reducing the duplication to the property binding in `getDerivedStateFromProps` in `Virtuoso` and `GroupedVirtuoso` - something which I am fine with.

Please let me know what you think about this - I am open to discussion, and a second pair of eyes on the codebase is eye-opening. Note: I am also trying to keep the streams protocols internal to the store, with the intent of migrating them to the faster and smaller callbag library family. 

Next - the current change will call the listScanner twice, which is the most expensive operation in the store (querying the binary trees in `offsetList`). That's something which I would like to avoid. 

I see two possible approaches here: 
Reuse the `list$` range, which will include the overscanned items. My observation in the test cases I have is that the overscan is best kept to `0` anyway; doing so distributes the updates uniformly when the user scrolls through the list, rather than having a 5fps bump every now and then. That's the simplest (though not perfect) approach - I am not sure if you are using overscan in your current setup.

Or: create a derived stream with something like `combineLatest(scrollTop$, list$, viewportHeight$).pipe(map(...))` which filters out the only the visible items. Come to think of it, this might be relatively easy, but a tad more expensive than the first option. 

Please let me know what you think or if you need any help or clarification with any of the above.
 
Cheers,

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-virtuoso",
-  "version": "0.10.2",
```
joshacheson: Suggested version bump because a feature is being added

playground/index.tsx
joshacheson: @petyosi it occurs to me leaving this in might make sense, since its just a playground, but it makes the console pretty noisy. Let me know what you prefer.

src/VirtuosoStore.tsx
```diff
@@ -240,6 +240,7 @@ const VirtuosoStore = ({ overscan = 0, totalCount = 0, itemHeight }: TVirtuosoCo
     initialItemCount: makeInput(initialItemCount$),
 
     list: makeOutput(list$),
+    itemsRendered: makeOutput(list$),
```
joshacheson: I decided to re-use `list$` straightforwardly. Some thoughts

1) When making my first attempt, the cost of `listScanner` had occurred to me, and I wondered if you would bring up running it twice. And so fair enough, I am definitely inclined to go with the cheapest option, re-using `list$`.
2) In my use cases I do not use `overscan`. I was trying to make this feature work "as expected" for other users, but of course what probably makes sense is to do `itemsRendered` and if something like `itemsInView` is actually needed by people in the future (because they're using overscan) we can tackle that later.

joshacheson:
@petyosi I think this is good for another review. Also, please let me know if you have preferences for commit message style etc when that time comes.

petyosi:
@joshacheson Thanks! I will review it later today and get back to you.

petyosi:
@joshacheson Thank you very much for your contribution! Merged the proposed changes and released v0.11.0. On to updating the docs. 

Cheers,

joshacheson:
heya @petyosi, it looks like the published `v0.11.0` does not include these changes. Any chance we can publish a `v0.12.0` that includes them?

petyosi:
I think I managed to somehow screw up the contents of 0.11.0 - another issue was raised for that. Can you check v0.11.1? The code seems to [be there](https://unpkg.com/browse/react-virtuoso@0.11.1/dist/react-virtuoso.cjs.development.js).

joshacheson:
@petyosi looks good. Could a simple catch here be to add a `"prepublishOnly": "npm run build"` entry to scripts?

petyosi:
🙇 Thanks for pointing me to that.

Issue #29, opened 2019-08-14
Create callback prop for getting total height


theonlylies:
I need get full height for implement my own scroll behavior around `scrollTop` for `VirtuosoScroller`.
I think that need/can to be implemented some callback prop for `Virtuoso`, that can be passed into `VirtuosoView` and call function with `fillerHeight` as param.
if you know some better solution. pls tell me 
if this idea not bad - i can try to create PR
sorry for my lang

petyosi:
Hey, 

Thank you very much for your suggestion. You can expose the fillerHeight via callback prop easily. The `isScrolling` property is exposed in a similar manner. You don't need to change the VirtuosoView. Check https://github.com/petyosi/react-virtuoso/pull/28 request, which does something similar (exposing output stream as a prop).

Let me know if you need any help with the PR. 
Cheers,

theonlylies:
@petyosi 
i add some dirty PR. I'l be glad if you help me with this #30

PR #30, opened 2019-08-18
height observer for #29 issue

theonlylies:
This code working, but i have some issue.
This callback get new height before DOM calc new scrollHeight for ScrollContainer(i use my implementation for scrollcontainer cause i want access to scrollable div)
If you have any suggestion pls tell me.
Maybe this feature must be implemented as ResizeObserver over VirtuosoFiller ref !?

### DIFF
--- a/playground/index.tsx
+++ b/playground/index.tsx
@@ -46,6 +46,9 @@ const App = () => {
           itemsRendered={items => {
             console.log({ itemsRendered: items })
           }}
+          heightObserver={height => {
+            console.log('List height:' + height)
+          }}
           ItemContainer={React.memo(ItemContainer)}
           ListContainer={React.memo(ListContainer)}
           GroupContainer={React.memo(GroupContainer)}
--- a/src/GroupedVirtuoso.tsx
+++ b/src/GroupedVirtuoso.tsx
@@ -24,6 +24,7 @@ export class GroupedVirtuoso extends PureComponent<GroupedVirtuosoProps, Virtuos
     state.groupCounts(props.groupCounts)
     state.groupIndices(props.groupIndices)
     state.itemsRendered(props.itemsRendered)
+    state.heightObserver(props.heightObserver)
     return null
   }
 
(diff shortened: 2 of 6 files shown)


petyosi:
The code looks good. Can you reproduce what's going on on your side (including the scroll container) in a stackblitz sample? I will take a look.

theonlylies:
@petyosi i add new commit to PR.
It's part of code that demonstrate behaviors of 2 ways:
`
const heightObserverUseOutput = (height) => {
        if (list.current)
            console.log("heightObserverUseOutput scrollHeight:" + list.current.target.scrollHeight);
    };

    const heightObserverUseHeight = (height) => {
        if (list.current)
            console.log("heightObserverUseHeight scrollHeight:" + list.current.target.scrollHeight);
    };
....
<Virtuoso
            ScrollContainer={scroller.current}
            style={{width: '100%', height: '100%', marginTop: 10, marginBottom: 10}}
            overscan={0}
            totalCount={messages.length}
            item={rendererVirtuoso}
            ref={virtuoso}
            heightObserver={heightObserverUseOutput}
            heightObserverTest={heightObserverUseHeight}
/>
`
In this code list.current is ref, which stores ResizeObserverEntry for my own ScrollContainer(i use ResizeObserver on it cause i want to watch change of clientHeight) 
and i output like this:
`
heightObserverUseOutput scrollHeight:480
heightObserverUseOutput scrollHeight:480
heightObserverUseHeight scrollHeight:540
`
so, i can see that useOutput way trigger callback 2 times(thats bad for logic) and scrollHeight on ScrollContainer remains old(i think because useOutput give us calculated but not applied height), but in callback passed new height! Accordingly, we cannot scroll with scrollTop at the time of the call heightObserverUseOutput callback!!!
sorry for this dirty explanation, but i can’t understand how to use my own build react-virtuoso on StackBlitz

petyosi:
Hey, 

In case you have not resolved this, you can use the playground folder to show me what the problem is. It will also really help me if I can understand your end goal here. What is the behavior you are trying to implement? Any link to an example of similar behavior will help.

A clue towards what you describe (I am not sure I understand it well enough) - the total height gets changed a lot, and initially reports "odd" values, due to the initial measurements. 

Cheers,

PR #32, opened 2019-08-19
heightObserver feature

theonlylies:
I delete old PR cause i had my own mistake.This code work like a charm
If you want i can make PR to rep with docs


petyosi:
Great! Just one minor thing: I would like to have the property name to `totalListHeightChanged` so that it matches the `itemsRendered` callback. 

Thank you!

theonlylies:
> Great! Just one minor thing: I would like to have the property name to `totalListHeightChanged` so that it matches the `itemsRendered` callback.
> 
> Thank you!

Ok. i renamed prop.
Thank you for awesome lib!

petyosi:
Thanks! I will update the documentation and publish a new version over the weekend.

PR #77, opened 2020-03-09
Remove inline style tag in favour of inline styled components

pedroalves0:
This PR is motivated by `VirtuosoStyle` injecting a style tag into the document's head, meaning that it collides with the recommended CSP guidelines, to remove any inlined style tags.

The sticky elements style attributes will now be passed directly to the component, not affecting possibly existing CSP rules.

### DIFF
--- a/src/Style.tsx
+++ b/src/Style.tsx
@@ -1,44 +1,4 @@
-import { useLayoutEffect, useRef, FC, CSSProperties } from 'react'
-
-export const VirtuosoStyle: FC<{
-  stickyClassName: string
-}> = ({ stickyClassName }) => {
-  const style = useRef<HTMLStyleElement | null>(null)
-  useLayoutEffect(() => {
-    const styleEl = document.createElement('style')
-    document.head.appendChild(styleEl)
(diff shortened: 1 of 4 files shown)


petyosi:
Cool PR! I will review and if everything works fine, will merge.

petyosi:
@pedroalves0 I like the approach, but is it possible to move the test to a singleton of some sort? Right now, each mount of the list will cause the re-creation of the test element. Probably a cheap operation, but still something I would like to avoid.

pedroalves0:
> Probably a cheap operation, but still something I would like to avoid.

Yeah, it's not a heavy computation, but I understand your concern.

> @pedroalves0 I like the approach, but is it possible to move the test to a singleton of some sort?

Done, I've moved the test logic to a memoized function, it's now being called only once.

petyosi:
Clever use of memoize! 
I am reluctant to bring lodash as a dependency, though. I will tweak your implementation a bit (by inlining a simple memoize with no args, etc) and merge later this week.

Thank you!

petyosi:
This change is now available in v0.14.0

PR #79, opened 2020-03-10
feat(endReached): inverted scrolling

### LINKED ISSUE #80: Inverted scrolling
Creating this ticket to match the PR.

I would like to be able to call the `endReached` callback when I scroll `top`, meaning the callback should be fired when I get the `index` 0.

lbenie:
I'm trying a scrolling chat like behaviour where when we reach the `top` it triggers the `endReached` callback. 

So basically, bottom --> up instead of the current implementation top --> bottom.

I added a debounce in the `listEngine` to prevent multiple calls to `endReached`

closes #80

### DIFF
--- a/playground/grouped-inverted.tsx
+++ b/playground/grouped-inverted.tsx
@@ -1,0 +1,61 @@
+import { generateGroupedUsers } from './fakeData'
+import React, { useMemo, useState, useRef, useCallback, useEffect } from 'react'
+import { render } from 'react-dom'
+import { GroupedVirtuoso } from '../src'
+
+// Slices the total groups to the groups
+// which contain the items so far
+// for example, if you have [10, 10, 10, 10]
+// groups in total, slicing them to 23 will result in [10, 10, 3]
+const calculateGroupsSoFar = (totalGroups: Array<any>, count: number) => {
+  const groups = []
+  let i = 0
+  do {
+    const group = totalGroups[i]
+    groups.push(Math.min(group, count))
+    count -= group
+    i++
+  } while (count > 0 && i <= totalGroups.length)
+  return groups
+}
+
+const App = () => {
+  const { groups, groupCounts } = useMemo(() => generateGroupedUsers(500), [])
+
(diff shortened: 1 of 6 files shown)


playground/grouped-inverted.tsx
```diff
@@ -0,0 +1,61 @@
...
+      style={{ height: '350px', width: '400px' }}
+      groupCounts={currentGroupCounts}
+      group={(index: number) => <div>Group {groups[index]}</div>}
+      item={(index: number) => <div>User {index}</div>}
+      initialTopMostItemIndex={45}
+      endReached={index => {
+        console.warn('index', index)
+        loadMore()
+      }}
+      inverted={true}
```
ysael: [discussion]: I am wondering if instead of doing `inverted` we should create `startReached ` ?

petyosi:
@ysael @lbenie  I think startReached would be a better choice, indeed. An inverted flag sounds like it will offer much more than that (like starting the list from the bottom, etc.).

Question - why don't you use the the [rangeChange event](https://virtuoso.dev/range-change-callback/)? You can check the startIndex for zero. I might be missing something, though. 

Thank you both for the participation!

lbenie:
> @ysael @lbenie I think startReached would be a better choice, indeed. An inverted flag sounds like it will offer much more than that (like starting the list from the bottom, etc.).
> 
> Question - why don't you use the the [rangeChange event](https://virtuoso.dev/range-change-callback/)? You can check the startIndex for zero. I might be missing something, though.
> 
> Thank you both for the participation!

@petyosi Thanks for responding, after much more thoughts the `rangeChanged` event can work for this particular case. Although, the inverted list is something that could be useful for a consumer though.
What are your thoughts on this?

If @ysael PR is merged and released, we can continue our work 👍

petyosi:
I merged the PR in v0.14.0 - please give it a try. If it works for you, let me know. 

Cheers,

lbenie:
> I merged the PR in v0.14.0 - please give it a try. If it works for you, let me know.
> 
> Cheers,

Works thanks!

PR #84, opened 2020-03-17
feat(groupedvirtuoso): support at bottom state changed

lbenie:
For now we need to use the `rangedChanged` callback but this is better to check if we're at the bottom of the list :)


petyosi:
Thank you, I will publish a new version with the changes later today.

Issue #97, opened 2020-04-15
Issues when filtering a source (array of objects)


toonvd:
Hi

I might be approaching this incorrectly, but I am attempting to filter list items based on a condition.
When the amount of data gets smaller, the lib tries to render too much (more keys than the totalcount).

Is my approach here wrong?
https://stackblitz.com/edit/react-mjiema

toonvd:
Update:

The issue does not exist in v0.12.6.
This might be related to #87

petyosi:
Looks like a bug to me. I will leave a modified version [of the example](https://stackblitz.com/edit/react-mjiema-5mhnbp?file=example.js) here for further examination.

toonvd:
> Looks like a bug to me. I will leave a modified version [of the example](https://stackblitz.com/edit/react-mjiema-5mhnbp?file=example.js) here for further examination.

Hi, yes, I have started debugging in listEngine since it gives back an empty array after filtering.

toonvd:
Reminder for myself to narrow it down: broke in v0.13.4

Issue #91, opened 2020-04-13
initialTopMostItemIndex doesn't work properly in Chrome


lesha1201:
When I try to use `initialTopMostItemIndex`, the list doesn't render any items. I tried to open an example from official page. It works on https://virtuoso.dev/initial-index/ but it doesn't work when I open in StackBlitz. I decided to download and try to build it but it didn't work neither. Also I tried it in playground and it didn't work there too. But for some reason it works in Firefox.

![react-kn82nn-ohhw1t-StackBlitz-Google-Chrome-2020-04-13-10-27-06](https://user-images.githubusercontent.com/10157660/79101830-1c22ae00-7d72-11ea-83fd-15895a9adbc3.gif)

By the way it works when `initialTopMostItemIndex` is either first element or last

**Environment:**
OS: Windows 10 1909
Browser: Chrome 80 (tried with disabled extensions), Chrome 81 mobile (Android)
Virtuoso: 0.15

petyosi:
I just [forked the example from the page in StackBlitz](https://stackblitz.com/edit/react-l1eibb?file=example.js) and it works fine for me. Can you elaborate on any additional steps?

lesha1201:
@petyosi It happens for me without any additional steps. I thought that it can happen only for me but I also tried on mobile and it behaves the same. I just tried on mobile again and noticed that if I reload https://virtuoso.dev/initial-index/ sometimes the list disappears even there. Can you try https://virtuoso.dev/initial-index/ to open it on mobile (Chrome), scroll to bottom to see an example and reload several times?

I will try to reproduce it on my laptop (linux) later and will report once I do this

lesha1201:
Verified the same behavior on another device. It happens periodically
Video: https://www.dropbox.com/s/4foqhy0hu9sqttb/2020-04-13%2013-29-43.mkv?dl=0

petyosi:
This looks like a rendering glitch - the scrollbar is there, but the items are not visible. Can you inspect the elements? Also, What's this - Chrome OS or Linux?

lesha1201:
@petyosi There is no elements. If I use footer then there is only the footer. It happens on Windows 10 (first video), Android, Linux (second video) in Chrome

lesha1201:
I tried to log `items` in `VirtuosoList` and here is the result:

![image](https://user-images.githubusercontent.com/10157660/79120791-6e78c480-7d9c-11ea-922a-dafc3c650741.png)

[...]

PR #99, opened 2020-04-18
Fix duplicate and inconsistent rendering sequence

petyosi:
One of the issues found with the problem is described in #97.
Probably many other reincarnations exist. 

The root cause of the issue was that prop updates (in this case, the `renderItem` prop) 
had higher priority compared to the observable updates. 
The solution was to move the render function calculation inside the store.

### DIFF
--- a/src/Virtuoso.tsx
+++ b/src/Virtuoso.tsx
@@ -1,28 +1,24 @@
 import React, {
-  ComponentType,
   CSSProperties,
   FC,
   forwardRef,
   ReactElement,
-  useCallback,
   useImperativeHandle,
   useLayoutEffect,
   useState,
 } from 'react'
 import { TScrollLocation } from './EngineCommons'
-import { ListRange, ScrollSeekToggle } from './engines/scrollSeekEngine'
+import { ListRange, ScrollSeekConfiguration } from './engines/scrollSeekEngine'
 import { ListItem } from './GroupIndexTransposer'
 import { TSubscriber } from './tinyrx'
 import { VirtuosoContext } from './VirtuosoContext'
-import { TRender, TRenderProps } from './VirtuosoList'
+import { TRenderProps } from './VirtuosoList'
 import { VirtuosoStore } from './VirtuosoStore'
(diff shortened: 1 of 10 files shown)


src/Virtuoso.tsx
```diff
@@ -98,7 +87,7 @@ export interface VirtuosoMethods {
 }
 
 export const Virtuoso = forwardRef<VirtuosoMethods, VirtuosoProps>((props, ref) => {
-  const [state] = useState(VirtuosoStore(props))
+  const [state] = useState(() => VirtuosoStore(props))
```
petyosi, line 90: @toonvd - this was the reason for the empty list you were observing. I am the noob who does not know how to use hooks in this case; With each re-render a new store instance was created, only to be immediately discarded. These are the specifics of the `useState` hook.

However, this was not the real problem. You can review the rest of the changes - and if you wish, please test the branch on your side.

toonvd: Since we both come from classes, I think this was pretty hard to detect xD I have a few hours tommorow for thorough testing!

petyosi: Can you please comment here when done? If not, I am publishing this on Monday, collected a few more from @lesha1201 in this release.

toonvd: Absolutely. I'll try to do it in the morning!

toonvd: Hi @petyosi, first tests with playground are going well. I notice a performance increase and no more errors. #91 seems to be fixed for me also. Is there any way to get this branch in one of my projects to doublecheck?

petyosi: Yes @toonvd . Run `yarn build`, and `yarn pack`. You will get a tarball file in the root dir. After that, from your project, run `npm i ../path/to/virtuoso/tarball.tgz` (fix the exact file name).

toonvd: Hi, selenium tests passed and some manual tests also on that particular project.
I think this should be ok for a new version.

PR #113, opened 2020-05-15
feat(grid-virtuoso): add `initialItemCount` prop

mohtada-h:
This PR adds `initialItemCount` prop to `VirtuosoGrid` and use in `VirtuosoGridEngine` for initializing item range(useful for SSR).


petyosi:
Available in v0.16.6

PR #124, opened 2020-06-07
use `transform: translateY` instead of `marginTop` in grid

mohtada-h:


petyosi:
Hey, @mohtada-h - I think I used `translateY` at some point and it caused some problems, but I can't remember what. Why is this suggestion?

mohtada-h:
Hey, @petyosi. I think this change improves performance. You can check these links for more information:
https://developers.google.com/web/fundamentals/performance/rendering/
https://csstriggers.com/margin-top
https://csstriggers.com/transform

petyosi:
I think there were some rendering glitches with transformY, so I will check this.

exentrich:
@petyosi maybe make same improvement for List? I checked locally and it works without problem

PR #136, opened 2020-06-12
Performance improvement in grid

mohtada-h:
This PR contains two improvements:

1. Add `computeItemKey` to the grid props
1.  Replace `transform` and filler component with `padding-top` and `padding-bottom`.

This technique is used in the implementation of virtual scrolling on Instagram. Based on my test changing `transform` will cause layout shifting. Also, the filler component was not necessary because we can update the `padding-bottom` to fill the remaining space(update style of one component instead of two components).


petyosi:
Thank you very much for that, will play with it in the next couple of days before merging. The padding approach is an interesting one - do you by any chance have some perf benchmarks? I might as well consider this for the list itself. 

Regards,

PR #147, opened 2020-07-23
Make it possible to override tabIndex

exentrich:
In my case I don't want to give scroller a tab focus, but right now I can't override this behaviour

### DIFF
--- a/src/VirtuosoScroller.tsx
+++ b/src/VirtuosoScroller.tsx
@@ -15,7 +15,7 @@ export type TScrollContainer = FC<{
   scrollTo: (callback: (scrollTop: ScrollToOptions) => void) => void
 }>
 
-const DefaultScrollContainer: TScrollContainer = ({ className, style, reportScrollTop, scrollTo, children }) => {
+const DefaultScrollContainer: TScrollContainer = ({ className, style, tabIndex='0', reportScrollTop, scrollTo, children }) => {
   const elRef = useRef<HTMLElement | null>(null)
   const smoothScrollTarget = useRef<number | null>(null)
   const currentScrollTop = useRef<number | null>()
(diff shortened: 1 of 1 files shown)


petyosi:
Hey! 

Question - why do you need to disable that? The tab index value helps with keyboard navigation - people can click on the list and scroll with the arrow keys. 

Also, I think that you can override the scroll container component and remove that attribute. 

And finally,  the PR seems incomplete because there is no way to pass a prop for the tab index from the top level component :)

Cheers!

petyosi:
@exentrich, bump, did you get to my questions?

Regards,

abhishiv:
Hey @petyosi I'm here with same issue. It's great for keyboard navigation, but some people might be using their own spatial navigation infrastructure insted(like for tv apps for example) and that might not like this prop.

That's my case as well.

petyosi:
Hey @abhishiv @exentrich - here is [how you can remove tab index](https://stackblitz.com/edit/react-drykey?file=example.js) by passing a custom scroll container. 

@abhishiv - I would love to see the component on a tv app :). Do you have any demo or project I can look at? Cheers!

exentrich:
@petyosi Items itself in my virtual list have `tabIndex`'s, so removing tabIndex from list have no consequences for accessibility.
Also I don't see any use-case for my app why user want to focus entire list, because my list already support arrow key navigation (custom implementation that highlight "active" item)

Custom scroll container is a solution, but for my opinion too heavy for such small change :)
In my opinion better if Virtuoso component itself will have `tabIndex` property.

petyosi:
Something which I am working on is to transpose all standard HTML attributes of the component to the root node (the scroller). This would resolve the problem in an elegant manner.

petyosi:
@exentrich v1 passes through all DOM properties to the outermost element. Check #203 for installation instructions.

PR #149, opened 2020-07-25
Add `header` prop

shoNagai:
This PR adds a header prop to Virtuoso, GroupedVirtuoso to help show indicators during pagination when using the list from bottom to top.

I'm creating a chat list like slack. and using scrolling from bottom to top and the indicator should be at the top of the list.
I added the header in the same way as the footer.


petyosi:
Thanks! I will review that thoroughly.

PR #150, opened 2020-07-25
Fix atBottom computation when content is smaller than the viewport.

gchallen:
As title. Very simple fix.


petyosi:
Thanks! Makes sense, I will batch that with a few other updates.

PR #156, opened 2020-08-08
Migrate react import

### LINKED ISSUE #155: Migrate to import * as React from 'react' ?
First of all, thank you for the work.

As far as I know, `import React from 'react'` is not correct.
ref: https://github.com/facebook/react/pull/18102

there is no default export in react codebase https://github.com/facebook/react/blob/master/packages/react/src/React.js,  It just happened to work with bundlers.

I though we should stick to what react core recommend.

you use `import React from 'react'` in the codebase and `import * as React from 'react'` in the README.md

I was just wondering if there is any reason you don't go with the recommended way in the codebase?

kodai3:
migrate react import 
from 
```
import React from 'react'
```
to
```
import * as React from 'react'
```

Closes #155

### DIFF
--- a/playground/grid.tsx
+++ b/playground/grid.tsx
@@ -1,4 +1,5 @@
-import React, { useRef, useEffect, useCallback, useState } from 'react'
+import * as React from 'react'
+import { useRef, useEffect, useCallback, useState } from 'react'
 import * as ReactDOM from 'react-dom'
 
 import { VirtuosoGrid } from '../src/VirtuosoGrid'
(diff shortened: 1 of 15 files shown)


playground/grid.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { useRef, useEffect, useCallback, useState } from 'react'
+import * as React from 'react'
+import { useRef, useEffect, useCallback, useState } from 'react'
```
kodai3, line 2: so far, I just changed as react core did in their migration work.

Essentially, it is importing same thing twice.
Because react team leave as it is, I guess it is fine.
  but, my personal preference is like bellow.

```suggestion
import * as React from 'react'

const { useRef, useEffect, useCallback, useState } = React
```

kodai3: another option would be 
```suggestion
import * as React from 'react'
```
and use `React.x`

PR #159, opened 2020-08-09
Add `emptyComponent` prop

### LINKED ISSUE #142: Empty component prop
An `emptyComponent` option could be a nice addition to render a component when the list is empty (i.e. `totalCount` = 0). A bit like MUI's `noOptionsText` on `autocomplete`, which renders when the options count is 0.

Just started using this library by the way and it's been a complete pleasure so far on a bit of pain point (infinite scrollers). So kudos!

shoNagai:
This PR adds a emptyComponent prop to Virtuoso, GroupedVirtuoso.
The emptyComponent is displayed when the item size is 0.

Closes #142

### DIFF
--- a/src/VirtuosoView.tsx
+++ b/src/VirtuosoView.tsx
@@ -6,6 +6,7 @@ import { VirtuosoContext } from './VirtuosoContext'
 import { VirtuosoFiller } from './VirtuosoFiller'
 import { VirtuosoList } from './VirtuosoList'
 import { TScrollContainer, VirtuosoScroller } from './VirtuosoScroller'
+import { ListItem } from './GroupIndexTransposer'
 
 export const DefaultHeaderContainer: React.FC<{ headerRef: CallbackRef }> = ({ children, headerRef }) => (
   <header ref={headerRef}>{children}</header>
@@ -114,6 +115,7 @@ export const VirtuosoView: React.FC<{
   HeaderContainer?: THeaderContainer
   FooterContainer?: TFooterContainer
   fixedItemHeight: boolean
+  emptyComponent?: React.ReactNode
 }> = ({
   style,
   header,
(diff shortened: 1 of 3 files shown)


src/VirtuosoView.tsx
```diff
@@ -144,7 +148,7 @@ export const VirtuosoView: React.FC<{
       <div ref={viewportCallbackRef} style={viewportStyle}>
         <ListWrapper fixedItemHeight={fixedItemHeight} ListContainer={ListContainer}>
           {header && <VirtuosoHeader header={header} HeaderContainer={HeaderContainer} />}
-          <VirtuosoList />
+          {items.length === 0 ? <>{emptyComponent}</> : <VirtuosoList />}
```
petyosi, line 151: I think that grabbing the items in the view will cause an additional render cycle of the View. 

Can you please move this check in the `VirtuosoList`? I know that this means one more prop drill (working to get that resolved, btw). Thanks!

shoNagai: Thank you for check pull request.
It's exactly as you said.
I corrected the mistake.

petyosi:
Hey @shoNagai , 

I merged the capability in `v0.20.0` but applied a modification - the property is of `ComponentType`, so that it stays consistent with the rest of the component properties. This is more flexible IMO, since the contents of the empty component do not have to be pre-rendered. 

Let me know what you think about it.

shoNagai:
> Hey @shoNagai ,
> 
> I merged the capability in `v0.20.0` but applied a modification - the property is of `ComponentType`, so that it stays consistent with the rest of the component properties. This is more flexible IMO, since the contents of the empty component do not have to be pre-rendered.
> 
> Let me know what you think about it.

Thanks for the modification.
I think ComponentType is fine, and it's good to be consistent with the others.

Issue #157, opened 2020-08-08
Remove default style height


gmattie:
I believe that it would be ideal to remove the default style height of `40rem`.  Doing so would avoid the need to declare inline styling for a `<Virtuoso>` component in favor of unopinionated style management via external style sheets or styled components (or inline styling).  Additionally, it would remove arbitrary code.

For example:  I would like to use Sass to assign the height of the list and display a custom scroll bar.

**Sass:**
```
@mixin scrollbar {
    // scrollbar custom design
}

.virtuosoList {
    @include scrollbar;
    height: 100%;
}
```
**React:**
```
<Virtuoso
    className={"virtuosoList"}
    totalCount={200}
    item={index => <div>Item {index}</div>}
/>
```

However, the `40rem` of the default `style` height will override my `className` height.  While I *could* mark my Sass height as `!important` so that it is not overridden, any use of `!important` outside of testing or debugging is both super ghetto and problematic.

Another potential option could be to simply replace the default style height of `40rem` with `100%`.  While that would still override a `className` height, declaring the height of a `<Virtuoso>` component could be optional and/or managed by the component's container.

petyosi:
Thanks, I think 100% percent would be a better and sensible choice indeed.

petyosi:
Available in `v0.20.0`

PR #160, opened 2020-08-10
Replaced default style height of "40rem" with "100%"

gmattie:
Replaces the default style height of "40rem" with "100%".

Resolves:  #157


petyosi:
Merged in `v0.20.0`. Thank you @gmattie !

PR #164, opened 2020-08-14
VirtuosoGrid | User's scroll behavior

KiyoshiMori:
* added playground/grid-smooth-behavior.html/tsx
* changed VirtuosoGridEngine/scrollTo$ return value

PS: Are there were some problems with non-auto behavior or was it just forgotten? When I tested it, I hadn't found any problems.

### DIFF
--- a/playground/grid-smooth-behavior.html
+++ b/playground/grid-smooth-behavior.html
@@ -1,0 +1,14 @@
+<html lang="en">
+<head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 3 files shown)


petyosi:
@KiyoshiMori Thanks! I will review your other PR and merge those together. 

`behavior: 'smooth'` is tricky; the implementation is not bullet-proof. There are certain cases where the result location won't be accurate - for example, if the items are complex and load dynamically sized images, etc. However, it will work and look great for simpler lists.

petyosi:
Rebased and merged in `v0.20.0`

PR #167, opened 2020-08-18
VirtuosoGrid | Offsets fix

KiyoshiMori:
* merge top and bottom padding values to one object-value at grid engine

Since listOffset$ and remainingHeight$ were updated separately, it caused two rerenders, which was making your scroll position have a "shaking" effect. With this commit, they are updating synchronously, and it fixes the "shaking" problem.
PS: you can check that shaking effect by changing the item height at playground/grid.tsx to 500px in the current master-branch.


petyosi:
@KiyoshiMori thank you for this PR, and I am sorry for delaying the review. Can you explain why is `isInit$` necessary?

petyosi:
This is fixed as part of the rewrite in v1. Please check #203 for instructions. There are breaking changes as well.

PR #185, opened 2020-10-27
changed scrolled to Bottom calc to be less then 1

### LINKED ISSUE #184: atBottomStateChange always return false when browser zoom is set to 90%
the following is a screen shot of dev tools showing the issue. we simply placed a breakpoint in the scrolledToBottomEngine function to see the values. please look at the watch list for clarification. looking at the scrollTop variable a decimal value simply does not make any sense to me.

image

ZNackasha:
this potentially fixes #184


petyosi:
@ZNackasha  - I see #189 as an alternative, will review both PRs.

petyosi:
@ZNackasha thanks for the idea on how to fix that. I merged the #189 since it includes a test case.

PR #189, opened 2020-11-02
fix scrolledToBottom calculation

ath92:
This fixes the `scrolledToBottom` calculation so that it also works when window is zoomed in/out. Zooming introduces rounding issues, which this PR fixes. This is related to https://github.com/petyosi/react-virtuoso/issues/184.

Note that the original issue can easily be reproduced by going to https://virtuoso.dev/stick-to-bottom/ and changing the zoom level of the browser. 

This is an alternative fix to https://github.com/petyosi/react-virtuoso/pull/185, which was causing this function to return `true` even when not scrolled to the bottom for me.

I also added the playground files I used to test this change locally. Let me know if they should be removed for this PR to be merged. You can verify locally what this PR does by simply removing the change in `scrolledToBottomEngine.ts`; this should break the behaviour in the playground example.

### DIFF
--- a/playground/followBottom.html
+++ b/playground/followBottom.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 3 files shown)


petyosi:
@ath92 @mahboubii - thanks for that. I will review thoroughly by the end of the week and merge. 

Cheers!

Issue #233, opened 2020-12-14
Infer type of 2nd positional parameter in itemContent callback


bb:
Hi Petyo,

thanks for creating react-virtuoso. It works really great for me.

I just upgraded to 1.0 and wanted to try the new `data` parameter in `itemContent` and noticed it was typed `any` instead of the inferable type from `<Virtuoso data={...} ...` prop.

I think it can be added like this:

```ts
export interface VirtuosoProps<ItemType extends any> extends Omit<ListProps, 'groupCounts' | 'groupContent'> {
...
  data?: ItemType[]
...
  itemContent?: (index: number, data?: ItemType) => ReactNode
...

export interface GroupedVirtuosoProps<ItemType extends any>
  extends Omit<VirtuosoProps<ItemType>, 'totalCount' | 'itemContent'>,
...
```

However, I don't know to parameterize the `export const` lines at the very end of components.tsx, yet. Maybe you do? If you have a pointer how to fix that, I might be able to provide a PR.

petyosi:
@bb - Me neither, TBH. I googled several "generic React components" phrases, but nothing seems to make sense.

bb:
Thanks for the quick reply! I don't have the time to investigate deeper right now, but I'll give it another try later. Or maybe somebody else chimes in. 😉

bb:
For Future-Me or whoever implements that: I just noticed that this would ideally also apply to `data` property of `Item` and `RecordItem` as well as `ItemContent` and `GroupItemContent` (defined in interfaces.ts). This would help e.g. in `itemsRendered`.

petyosi:
I don't think (but I might be wrong) that you have to interfere with interfaces.ts. The "official" typings of the component are re-declared in `components.ts` for documentation purposes. While they do extend some interfaces from interfaces.ts, this is mostly to avoid duplication.

seahorsepip:
I had a try at adding generic typing to the `Virtuoso` and `GroupedVirtuoso` components.

PR #247, opened 2020-12-23
Infer typing from data prop (if provided, else fallback to any)

seahorsepip:
Related #233


seahorsepip:
This PR also extends the props from `div`.

More info on type interference:
https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref
https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase

petyosi:
@seahorsepip thanks for cracking that! I cleaned up the lint and typing errors for tests in the push above.

bb:
Nice. Thanks!

PR #255, opened 2021-01-08
fix: allow passing readonly array data

n1ru4l:


petyosi:
Fixed in 31a71babea3addde9cf4d0f00ac4f21f4e0c0ea4, thank you!

PR #258, opened 2021-01-12
Fix incorrect property name

mesmeri:
Looks like the property name 'itemCount' was added mistakenly and the information in the doc actually refers to 'itemContent'.


petyosi:
Thanks! Can you reword the commit to `docs(): fix incorrect property name` please?

PR #296, opened 2021-02-16
fix: Window Scrolling

### LINKED ISSUE #295: [BUG] Window scrolling not working
**Describe the bug**

New to this. Trying to use Window Scrolling (which is what drew me to this library, other than dynamic height of course!). 

It seems the list isn't updated (items below-the-fold don't show up)

[GIF](https://gfycat.com/vacantanyalaskanmalamute)

**Reproduction**

https://github.com/laggingreflex/repro-react-virtuoso-window-scrolling

I created that simple repro first. But then I tried to recreate it on codesandbox where the issue disappears! https://codesandbox.io/s/repro-react-virtuoso-window-scrolling-xstbt 

Not sure why. I can definitely repro the issue on a standalone app, but not on codesandbox. 

**To Reproduce**

For standalone app, clone the above repo, run it locally. 

Then just scroll. 

**Expected behavior**

It should show all 1000 elements

**Screenshots**

See above

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Tested both on Firefox & Chrome (same results)

laggingreflex:
Digging around to solve #295 I found that `document.documentElement.scrollTop` stayed 0 when I scrolled. So taking advice form [this][1] I included other alternatives to get the scroll value which seems to make it work! 

Fixes #295

[1]: https://stackoverflow.com/questions/20514596/document-documentelement-scrolltop-return-value-differs-in-chrome/33462363#33462363

### DIFF
--- a/src/hooks/useScrollTop.ts
+++ b/src/hooks/useScrollTop.ts
@@ -16,7 +16,12 @@ export default function useScrollTop(
   const handler = useCallback(
     (ev: Event) => {
       const el = ev.target as HTMLElement
-      const scrollTop = (el as any) === window || (el as any) === document ? document.documentElement.scrollTop : el.scrollTop
+      const scrollTop =
+        (el as any) === window || (el as any) === document
+          ? window.scrollY ||
+            window.pageYOffset ||
+            document.body.scrollTop + ((document.documentElement && document.documentElement.scrollTop) || 0)
+          : el.scrollTop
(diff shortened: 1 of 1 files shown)


petyosi:
Thank you, I will look into this shortly.

laggingreflex:
Nevermind, found a fix. I was missing `<!DOCTYPE html>` https://github.com/petyosi/react-virtuoso/issues/295#issuecomment-779633108

PR #328, opened 2021-04-01
fix(): window scroller works ios 12 and below

sei666:


petyosi:
Merged and released; the change is available in 1.6.0.

PR #335, opened 2021-04-10
feat(): support custom TopItemListComponent

ezsper:
This PR introduces a small new feature that allows the definition of a custom component for `TopItemListContainer`, this is useful to set a custom z-index or a different strategy for sticky position.

I had to do the following additional changes due to problems with setup:

Node: v12.18.3
NPM: 6.14.6
ProductName:    Mac OS X
ProductVersion: 10.15.7
BuildVersion:   19H2

#### 1)
I had to change this
```
const ListContainer: GridComponents['List'] = styled.div`
  display: flex;
  flex-wrap: wrap;
`
```
to 
```
const ListContainer = styled.div`
  display: flex;
  flex-wrap: wrap;
` as GridComponents['List']
```
because I was getting that `Types of property 'propTypes' are incompatible.` between `StyledComponent` and `GridProps`

#### 2)
I've also had to change this
```
 const observer = new ResizeObserver((entries) => {
    const element = entries[0].target as HTMLElement
```

to 

```
 const observer = new ResizeObserver((entries: ResizeObserverEntry[]) => {
    const element = entries[0].target as HTMLElement
```

explicitly assigning a type to `entries` as this was preventing me to do an initial `npm install`. As I was getting:
```
(rpt2 plugin) Error: src/hooks/useSize.ts(8,40): semantic error TS7006: Parameter 'entries' implicitly has an 'any' type.
```

### DIFF
--- a/e2e/grid-optimize-rendering.tsx
+++ b/e2e/grid-optimize-rendering.tsx
@@ -35,10 +35,10 @@ const ItemWrapper = styled.div`
   }
 `
 
-const ListContainer: GridComponents['List'] = styled.div`
+const ListContainer = styled.div`
   display: flex;
   flex-wrap: wrap;
-`
+` as GridComponents['List']
 
 const Item = React.memo<any>(({ item }: { item: { index: number; selected: boolean } }) => {
   console.log(`rendering Item ${item.index}`)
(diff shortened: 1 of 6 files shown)


petyosi:
Thanks! I recently tried to tighten up typechecking on the examples. Wonder why the fixes you had to do did not manifest in the CI 🤔.

PR #376, opened 2021-05-31
Add List.itemSize prop.

### LINKED ISSUE #375: Customize the item height lookup
I have react-virtuoso's `List` ~90% rendering a CSS grid-based table, but have hit one snag.

Per the customized rendering guide, I'm providing my own `List` component, which sets `display: grid` and `gridTemplateColumns`.

Normally with CSS grid, the cells are supposed to be immediate children of the `display: grid`.

For react-virtuoso, the `Item` wrapper `div` "gets in the way" between the `Item` and the `itemContent` return value ... however! a known/kosher way around this is to set the `Item`'s div as `display: contents`. This kinda "skips" the Item div in the CSS grid's logic, and makes the divs within the `Item` render as rows / cells really well (each of the rows here comes from `itemContent` and the column widths are set by the `gridTemplateColumn: auto auto auto` on the `List`):

image

The biggest gotcha is that `display: contents` also makes the `offsetHeight` of the `Item` zero (which makes sense, it's getting "skipped" in the DOM), which breaks automatic sizing.

Somewhat sneakily, I can "fix" automatic sizing by changing this line:

```
    const index = parseInt(child.dataset.index!)
    const knownSize = parseInt(child.dataset.knownSize!)
    // Special change here
    const size = (child.firstElementChild! as HTMLElement)[field]
```

To look at the `Item`'s child element to get the size.

So, the DOM ends up looking like:

* `List` item with `display: grid` set
  * `Item` item with `display: contents` set
    * List of cells for row i

And I'm "skipping" the 2nd level for size measurement, and measuring the cell height.

I know this is somewhat esoteric, but I assert being able to use react-virtuoso to virtualize CSS grids (i.e. without the limitations of `Grid`) would be pretty great/potentially widely used.

Do you have any thoughts on a `List` prop that would let me customize that `const size = child[field]` line to be a function that I could pass in?

So far I'm just running a locally built version and could potentially run this as a fork if we need to, but would be great to have this supported in a kosher way.

Also, happy to learn if I'm missing something obvious about a better way to do CSS grid-based tables in react-virtuoso.    

Thanks!

stephenh:
Fixes #375.

### DIFF
--- a/src/List.tsx
+++ b/src/List.tsx
@@ -172,8 +172,9 @@ export const Items = React.memo(function VirtuosoItems({ showTopList = false }: 
   const itemContent = useEmitterValue('itemContent')
   const groupContent = useEmitterValue('groupContent')
   const trackItemSizes = useEmitterValue('trackItemSizes')
+  const itemSize = useEmitterValue('itemSize')
 
-  const ref = useChangedChildSizes(sizeRanges, trackItemSizes)
+  const ref = useChangedChildSizes(sizeRanges, itemSize, trackItemSizes)
   const EmptyPlaceholder = useEmitterValue('EmptyPlaceholder')
   const ScrollSeekPlaceholder = useEmitterValue('ScrollSeekPlaceholder') || DefaultScrollSeekPlaceholder
   const ListComponent = useEmitterValue('ListComponent')!
@@ -418,6 +419,7 @@ export const { Component: List, usePublisher, useEmitterValue, useEmitter } = sy
       computeItemKey: 'computeItemKey',
       defaultItemHeight: 'defaultItemHeight',
       fixedItemHeight: 'fixedItemHeight',
+      itemSize: 'itemSize',
       scrollSeekConfiguration: 'scrollSeekConfiguration',
       headerFooterTag: 'headerFooterTag',
       data: 'data',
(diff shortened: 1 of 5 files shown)


src/List.tsx
```diff
@@ -172,8 +172,9 @@ export const Items = React.memo(function VirtuosoItems({ showTopList = false }:
   const itemContent = useEmitterValue('itemContent')
   const groupContent = useEmitterValue('groupContent')
   const trackItemSizes = useEmitterValue('trackItemSizes')
+  const itemSize = useEmitterValue('itemSize')
```
stephenh, line 175: `itemSize` is not a valid param to `useEmitterValue`

petyosi: (On mobile) I think you got everything right, maybe itemSize needs to be listed in listSystem?

stephenh: Ah, good hint! I saw the destructing of:

```
export const listSystem = u.system(
  ([
    { totalCount, sizeRanges, fixedItemSize, defaultItemSize, trackItemSizes, itemSize, data, firstItemIndex, groupIndices },
```

And just made the connection that it's destructuring against the 1st argument of `u.tup` which is `sizeSystem`, so yep I'm allowed to be `itemSize` there and then export it.

I pushed a few more commits and looks like everything is wired up now.

PR #378, opened 2021-06-03
fix(): use document.documentElement.scrollHeight

hanger121212:
scrollToIndex works with html height:100%


petyosi:
Makes sense - can you change the commit message from `fix() ` to `fix(): ` so that the proper deployment is triggered?

hanger121212:
> Makes sense - can you change the commit message from `fix() ` to `fix(): ` so that the proper deployment is triggered?

ok commit message changed to fix():

PR #395, opened 2021-06-23
fix: remove unused react-app-polyfill dependency

iamstarkov:
I dont believe its used


petyosi:
Thanks, looks like you are right. It has been right there from the very beginning, might as well be something that tsdx generated. I will double-check.

iamstarkov:
thanks, I appreciate it

PR #419, opened 2021-07-26
fix: from visible range calculation

cryptoethic:
I have missed an edge case in #417

When header height is greater than item height from visible range is wrongly calculated.


petyosi:
Can you revert the e2e test case change? It showcases an important test.

Issue #418, opened 2021-07-25
[BUG] Scroll flickering with expandable cards


cryptoethic:
Hi!

I got something hard to reproduce which cause us a lot of pain.

https://codesandbox.io/s/react-virtuoso-reverse-scrolling-header-issue-forked-fj5ys?file=/src/App.js

Sometimes when i scroll up/down reflow occurs and my scroll position is moved from place where I have been frame before.

I have prepared sandbox with expandable cards (onClick), and this gif showing the issue (it does not happen all the time but is frequent enough to be painful)
![virtuoso flickering](https://user-images.githubusercontent.com/44468286/126914596-92135e55-7474-479f-85e6-ad296390aefe.gif)

Any idea what can cause this issue? I will appreciate any help ;-)

petyosi:
I am not sure how much this example follows your actual source code, but to optimize this scenario, you do need to apply proper memoization on a few levels, like the parameters of the Virtuoso component as well as the rendering of your items. Take a look at this fork for more details. 

https://codesandbox.io/s/react-virtuoso-reverse-scrolling-header-issue-forked-txuow?file=/src/Row.js:997-1116

petyosi:
Perhaps I am missing something - can you be more specific about what do you all a scroll flickering? There is some scroll repositioning due to item resize, and in reverse mode, this is not fully avoidable.

cryptoethic:
I couldn't find a specific case where scroll jump occurs, but what can be seen on the above gif is something like that:

1. I'm scrolling down and expanding cards
2. I'm collapsing bottom card while it is still fully in my view
3. I'm starting scrolling upwards and than jump occurs - I'm moved some cards to the bottom.

Unfortunately this does not occur every time, but is frequent enough to be uncomfortable.

cryptoethic:
And here is an example where I only expand cards and scroll jump occurs when i start scrolling upwards.
![virtuoso flickering 3](https://user-images.githubusercontent.com/44468286/126969391-2333425f-aafd-456e-975f-fd49982960a3.gif)

petyosi:
Which browser and OS is this? I am doing similar steps and it works on my side. Chrome, MacOS. 

https://user-images.githubusercontent.com/13347/127023686-8c4546c5-d4fd-4e87-8351-bfc133d23b09.mov

cryptoethic:
Hey, I'm using Windows 10 with latest Chrome.

[...]

PR #422, opened 2021-08-01
Fix upward scroll system flickering

cryptoethic:
@petyosi  I think I got it.

So the problem with #418 was that u.scan subsystem from upwardScrollFixSystem was executing only when scrolling upwards, therefore items that were expanded while scrolling down had incorrect sizes and offsets in prevItems array.

This solution replaces execution order.

Please review this PR ;)

### DIFF
--- a/e2e/expandable-cards.tsx
+++ b/e2e/expandable-cards.tsx
@@ -1,0 +1,57 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import * as React from 'react'
+import { useState } from 'react'
+import { Virtuoso } from '../src'
+
+const Row = (props: any) => {
+  const { rowIndex, expanded, setExpanded } = props
+  const [ex, setEx] = useState(expanded)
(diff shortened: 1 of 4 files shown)


petyosi:
That's a good catch, seems plausible. I will take a look

cryptoethic:
Thanks! I know you have a lot of work in your pipeline but if you could take a look soon I would be very grateful. This issue is blocking for our release and we would like to get it fixed as soon as possible.

If it is not a solution we would like to focus our time for finding it as soon as possible ;-)

petyosi:
So far this looks very very correct. I just need to double-check before I merge. If this is blocking your release, I may recommend you use your fork for this purpose.

cryptoethic:
I had a problem when installing fork as npm module and getting a code review might be important as I don't want to introduce new issues ;)

Take your time!

petyosi:
Congrats, I would say that you are the 2nd person apart from me who has done a non-trivial change in the urx-based core of the project :).

cryptoethic:
Thanks for your help during this issue! Without it I would be nowhere near.

It seems like scroll flickering problem is solved. What's more I got some basic knowledge about urx system that is behind react-virtuoso and as I said before - I'm quite impressed!

Let me know in case you would need any help in the future ;)

Issue #343, opened 2021-04-19
Omit resize observer polyfill from bundle


anilanar:
Use case:

We target modern browsers only so we don't need the polyfill.

petyosi:
Can you submit a PR for that? I doubt that the rest of the code will work on any of those browsers, to be honest.  

https://caniuse.com/resizeobserver

anilanar:
@petyosi This could be a breaking change tho. Perhaps the rest of the code works for those browsers after babelifycation? Or are you already aware of other DOM/ES apis that the codebase relies on that wouldn't work for, let's say, Safari 12.1 or Safari 13.0 which are quite "modern" actually.

petyosi:
@anilanar isn't this a general problem with those polyfills? I am not following the development of those closely, but I was hoping that a browserlist configuration could make a polyfill "tree-shake" unnecessary code.

anilanar:
The ecmascript version target: the consensus is to always  transpile library code to ES5.

For Ecmascript polyfills: The general consensus is to not babelify node_modules, so no, nothing would be polyfilled for libraries. In those cases, library authors are either choose not to use those APIs or they polyfill the API for small stuff. If there's something big (let's say, async iterators), they drop a note in README.

For DOM polyfills (resize/mutation observers, web components etc), the consensus is to always specify them in README as dependencies, except for small things like requestAnimationFrame which is small enough to polyfill.

I think for babeling, asking users to force include virtuoso in their babel process is viable.

Yeah, it's a lot, and considering that nobody complained about this before, you must have done some things right already. I'll check how virtuoso is built for release and comment further.

anilanar:
> I was hoping that a browserlist configuration could make a polyfill "tree-shake" unnecessary code.

No. Instead, babel preset env (and/or auto-polyfill babel plugins) read the browserslist and choose whether to include a polyfill or not, but that doesn't apply to DOM polyfills. It's only for Ecmascript polyfills.

So as a library author, the ideal solution is to not include any Ecmascript polyfills but let users babelify library code. But as you can see in the above comment, the consensus is currently different due to too many different opinions from community leaders/library authors.

petyosi:
@anilanar Thanks, I guess I hoped for too much. Up to your case, I assume you can use something like this: https://webpack.js.org/configuration/resolve/ to cut off the dependency?

petyosi:
Resolved in v2.0.0

PR #442, opened 2021-08-31
perf(deps): don't include resize-observer-polyfill in bundle

michael42:
Fixes https://github.com/petyosi/react-virtuoso/issues/343.

This removes `resize-observer-polyfill` from the production dependencies and moves it to `devDependencies`. This is required because the tests still need the `resize-observer-polyfill`.

Since this is a **semver-major breaking change**, I also added a new section to the `README.md`.

Unfortunately, TypeScript 4.1 doesn't include the `ResizeObserver` typings, so I included two declarations for it that can be removed when updating the TypeScript version (I tried but gave up when the e2e tests did no longer compile because of `rootDir` issues).

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,17 @@ If you need to support reordering, [check the React Sortable HOC example](//virt
 
 For in-depth documentation and live examples of the supported features and live demos, check the [documentation website](//virtuoso.dev).
 
+## Browser support
+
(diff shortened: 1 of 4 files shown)


petyosi:
Thanks for that @michael42,

Based on https://caniuse.com/resizeobserver, I think it's a good call to do a v2 release. I have a few other minor-yet-breaking changes which I am hesitant to include. 

 I would also like to bump TypeScript while at it. I Will not merge this immediately and see if I can upgrade it first.

petyosi:
Once again, thanks for the nudge here. I took a slightly different approach for the SSR, since it might actually cause real SSR rendering to be broken.

Issue #473, opened 2021-10-19
[Feature request] Add fallback of globalThis


nereuseng:
I found the globalThis was used in some place, like [here](https://github.com/petyosi/react-virtuoso/blob/ffede3ff0bc9a3e3f7fd3816897e5cb767e70aa8/src/loggerSystem.ts#L33), that will cause crash and the runtime error on some old browser, could we add a fallback of globalThis like Window?

Thanks!

nmgh1:
Hi,
Even I faced same issue, run time error for globalThis undefined.
I faced this on Safari 12 or lower

PR #503, opened 2021-11-18
fix: globalThis fallback

gustavo-depaula:
Hello!

Even after https://github.com/petyosi/react-virtuoso/commit/922a4681f9a34253eb1f61b6137e23fc4576db5d , I still had the same problem reported in https://github.com/petyosi/react-virtuoso/issues/473 in old browsers:
image

I was able to simulate locally by running `delete globalThis` in the console and going to a screen that uses `react-virtuoso`, then the application crashed, as reported by Sentry. This happens because the browser throws a `ReferenceError` if you try to evaluate a non-defined variable:
image

Therefore `globalThis || window` threw an error in older browsers.

I tried locally the same scenario with this fix I'm submitting and my application is not breaking anymore with it.

### DIFF
--- a/src/loggerSystem.ts
+++ b/src/loggerSystem.ts
@@ -5,6 +5,11 @@ declare namespace globalThis {
   let VIRTUOSO_LOG_LEVEL: LogLevel | undefined
 }
 
+// eslint-disable-next-line @typescript-eslint/no-namespace
+declare namespace window {
+  let VIRTUOSO_LOG_LEVEL: LogLevel | undefined
+}
+
 export enum LogLevel {
   DEBUG,
   INFO,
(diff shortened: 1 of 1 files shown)


src/loggerSystem.ts
```diff
@@ -26,11 +31,19 @@ const CONSOLE_METHOD_MAP = {
   [LogLevel.ERROR]: 'error',
 } as const
 
+const getGlobalThis = () => {
+  try {
+    return globalThis
+  } catch (_) {
+    return window
+  }
```
petyosi, line 39:
```suggestion
  return typeof globalThis === 'undefined' ? window : globalThis
```
Can you test that?

petyosi:
Question - what's the browser that fails at that?

gustavo-depaula:
@petyosi old chrome and samsung browsers: 
image

petyosi:
This looks good. If the suggested change works, I will merge it - trying to avoid the try/catch overhead. Thank you!

gustavo-depaula:
Ok! Trying it ASAP!

src/loggerSystem.ts
gustavo-depaula: it worked! just pushed it

gustavo-depaula:
@petyosi tested it, and it works! pushed the new commit here

petyosi (approved):

gustavo-depaula:
BTW @petyosi, thanks for `react-virtuoso`, it's truly a great library, and it helped us with some hard performance issues we were having!

Issue #320, opened 2021-03-24
[Feature Request] add scrollElement to useWindowScroll


acapro:
This is by far one of the best windowing tools I have tried, and would love to be able to use it, but there is one feature that is sadly still missing and I have not yet been able to find a work around this yet.

I have a more or less complex layout with multiple overflowing divs and scrollbars, and it would be an awesome feature if the useWindowScroll could be extended to support other wrapping scrollable (overflow) elements besides the window. 

In react-virtualized there is a WindowScroller component that does just this, it takes an html element parameter called scrollElement (https://github.com/bvaughn/react-virtualized/blob/master/docs/WindowScroller.md)

Here is what I am trying to do:
<img width="815" alt="Screenshot 2021-03-24 at 16 11 00" src="https://user-images.githubusercontent.com/4762057/112334187-8995c900-8cbb-11eb-916f-52553813ef37.png">

petyosi:
Hey, 

What you explain is fairly simple to achieve. You can take a look at the commit which introduced the window scrolling behavior and extend it further. I am happy to accept a PR for that.

maxhelsel:
I second this. My _only_ issue with Virtuoso is the fact it doesn’t have support for assigning a custom parent scroll element like React Virtualized WindowScroller.

My use case is within modals and containers that have other elements within them; RV WindowScroller makes it easy and convenient to inject a scrollElement ref and carry on. 

I forked this package and spent a few hours with it, but I’m not experienced whatsoever with packages and just couldn’t get a handle on everything going on within the src file. Unfortunately, I don’t have the time to familiarize myself enough to see it through. 

If @petyosi or anyone else capable of implementing this feature stumbles upon this, here’s the link to where the window scrolling feature was introduced to Virtuoso: [https://github.com/petyosi/react-virtuoso/compare/v1.2.7...v1.3.0](https://github.com/petyosi/react-virtuoso/compare/v1.2.7...v1.3.0). 

My thinking was to provide an optional scrollElement prop for Virtuoso component, to be swapped within buildWindowScroller if the scrollElement props exists.

- Uses window (via buildWindowScroller within src/List)
```
<Virtuoso 
  useWindowScroll={true} 
/>
```

- Uses custom scrollElement (via buildWindowScroller within src/List - swaps the customRef for the window). Figured it would be easiest just to try to slip the ref within the buildWindowScroller function and recycle all that logic.
```
<Virtuoso 
  useWindowScroll={true}
[...]

PR #556, opened 2022-02-03
Add scrollElement

cvanem:
PR for #320 

Adds a scrollElement input prop which mimics the useWindowScroll feature with the provided element.

See examples for usage:
scroll-element-grid-scroller
scroll-element-list-scroller


petyosi:
@cvanem this looks great - I will test it a bit more over the weekend but the implementation is correct overall.

petyosi:
Hey @cvanem, 

I tested the branch, and wanted to confirm something with you. The invisible items are rendered are due to the nested scrolling, right? 

image

Also, you can run `yarn lint --fix` and commit the changes? thanks!

cvanem:
@petyosi  

Yes, I believe that is correct, but it wouldn't hurt to have more eyes on it to make sure it is correct.

I originally tried implementing it with only the viewable boundary, but I found that the window contents would not refresh if another outer scroll div (not scrollElement) was adjusted to bring the list within the viewport.

I experimented a bit with adding an IntersectionObserver to trigger a refresh callback when this scenario occurred and did have some limited success but didn't really like the solution or use of thresholds and wasn't too confident with it.

I then implemented **getScrollElementInfo**, which seems to work so I went with that.  It would be good to have a few more eyes on that function to check it for accuracy.

maxhelsel:
@petyosi @cvanem Implementation looks good and is inline with what we all discussed in the original feature request thread!

A quick thought - I have not tested or played around with it, but when you state the following:

> I found that the window contents would not refresh if another outer scroll div (not scrollElement) was adjusted to bring the list within the viewport

Sounds like you're describing the Virtuoso component failing to update when dimensions/components change within the scroll element container? For example, you have a header or something that might render in and/or change size and skews the dimensions of the Virtuoso component relative to the viewport - therefore needing an update to recalculate dimensions.

The addition of **getScrollElementInfo** looks to be a fix, but I bring this up because I have been using React-Virtualized WindowScroller (RVWS) to allow for this entire ScrollElement use case in my virtualized lists, and that package has that same issue.

RVWS uses an outside method to address this very issue of updating dimensions, from the [documentation](https://github.com/bvaughn/react-virtualized/blob/master/docs/WindowScroller.md): 

> updatePosition
Recalculates scroll position from the top of page.
This method is automatically triggered when the component mounts as well as when the browser resizes. It should be manually called if the page header (eg any items in the DOM "above" the WindowScroller) resizes or changes.

When using RVWS I found myself having to call that method whenever additional components rendered into the scroll element container and pushed the List down.

Again, your **getScrollElementInfo** solution seems to address this on-the-fly instead having to manually call it. However, it might be worth a quick look to see what's going on in the [RVWS component code](https://github.com/bvaughn/react-virtualized/blob/master/source/WindowScroller/WindowScroller.js).

Otherwise, looks good to me. Hopefully I'll be able to drop React-Virtualized altogether soon!

src/hooks/useWindowViewportRect.ts
```diff
@@ -32,13 +45,15 @@ export default function useWindowViewportRectRef(callback: (info: WindowViewport
   }, [calculateInfo, ref])
 
   useEffect(() => {
-    window.addEventListener('scroll', windowEH)
-    window.addEventListener('resize', windowEH)
+    const element = scrollElement ? scrollElement : window
+
+    element?.addEventListener('scroll', windowEH)
+    element?.addEventListener('resize', windowEH)
```
petyosi, line 51: I just realized that this won't work for any element. Only the window has a resize event. We will need a resize observer here.

cvanem: Good catch, I will update.

cvanem: @petyosi I just pushed the resize observer if you can please review.  Thanks!

cvanem: One other thought I just had.  I currently disable the window resize event listener if a scrollElement is provided:

`!scrollElement && window.addEventListener('resize', windowEH)`

It looks like the resize observer is widely supported today so this is probably not a concern, but we could enable this as a fallback when scrollElement is provided and the browser doesn't support resize observers.  This would catch most of the scrollElement resize scenarios.

Any thoughts on the best/preferred approach when using the scrollElement prop?

1. Leave as is.  It looks like everything except internet explorer fully supports resize observers.
2. Always register the window resize event listener.  A few extra resize events shouldn't affect performance.
3. Register the window resize event listener as a fallback when a scrollElement is provided and typeof ResizeObserver === undefined.

cvanem: FYI, I did some more testing on my end and for the most part this PR works well, but I did see some issues where the row divs are not un-mounted when the list is scrolled from the bottom to above the top of the list so the list is out of view.   This looks like a pre-existing issue as far as I can tell as I could reproduce it on the window scroll example as well by adding a spacer above the list and scrolling down and then up.

I was able to get everything working as expected by making changes to the sizeRangeSystem -> visibleRange function, but I am not confident that these changes won't break something something else and it made a few of the automated tests fail....so I think it is best to evaluate this possible issue in separate PR/issue thread.

src/components.tsx
```diff
@@ -208,6 +208,11 @@ export interface VirtuosoProps<D> extends Omit<ListProps, 'groupCounts' | 'group
    */
   useWindowScroll?: boolean
 
+  /**
+   * Uses an external scroll element rather than wrapping the list in its own.
```
petyosi, line 212:
```suggestion
   * Pass a reference to a scrollable parent element, so that the list won't wrap in its own.
```
Assuming you are a native speaker, you can certainly improve on that. I would like to make it obvious that 
- a `ref` to an HTMLElement should be passed;
- the ref should point to a parent element (kind of common sense, but still).

src/hooks/useChangedChildSizes.ts
```diff
@@ -7,7 +7,8 @@ export default function useChangedListContentsSizes(
   itemSize: SizeFunction,
   enabled: boolean,
   scrollContainerStateCallback: (state: [number, number]) => void,
-  log: Log
+  log: Log,
+  scrollElement?: HTMLElement
```
petyosi, line 11: Looking for a better name here, as my brain blocks on `scrollElement` and `scrollableElement`. I would even go for `externalScrollableParentElement`?

petyosi:
@cvanem what do you think about renaming the prop to `externalScrollableParentElement`? I know it's longer, but we have code autocomplete these days.

src/hooks/useWindowViewportRect.ts
petyosi: I would say that the "proper" way to handle that would be the following:
- Use resize observer through the hook you made. It might not even be a hook so that it can be called conditionally (this part feels a bit weird right now). 
- The lib uses resize observers at a few other locations - refactor them to use the same code block.
- If `window.ResizeObserver` is not present, trigger an error, directing them towards [the polyfill](https://github.com/que-etc/resize-observer-polyfill)

I see that I have edit permissions on the PR, so I can do that as well so that we have the PR done properly.

cvanem:
> @cvanem what do you think about renaming the prop to `externalScrollableParentElement`? I know it's longer, but we have code autocomplete these days.

@petyosi I prefer short simple names but we can name it whatever you want.  The documentation can describe the usage in further detail if needed.

src/hooks/useChangedChildSizes.ts
cvanem: I like scrollElement.  Maybe someone else has a suggestion?

Maybe parentScrollElement, but that is starting to get long.

src/hooks/useWindowViewportRect.ts
cvanem: @petyosi Ok feel free to update as needed and let me know if you need me to do anything else to get this PR merged.

src/hooks/useChangedChildSizes.ts
maxhelsel: @petyosi @cvanem My favorite is **customScrollElement**, or maybe **customScrollParent**. Something like that. I think 3 words max, 4 seems like overkill

cvanem: @petyosa Yeah that would work.  My favorite is scrollHost or scrollElement.

src/hooks/useWindowViewportRect.ts
cvanem: @petyosi If you have a minute, could you please take look at the updated examples and code I tested in regards to my comment above about the sizeRangeSystem.  Here are the [changes](https://github.com/cvanem/react-virtuoso/commit/f4fd1eb4831a05ca40b53d4797f695cf77865131) I made to get everything working as expected.  Please take note of the changes in sizeRangeSystem.ts as this will affect/break other stuff, so no PR yet.

At this point, I am not sure what the correct solution would be without modifying the sizeRangeSystem, but I don't want to do that as it puts the architecture at risk of breaking something else.  

The size range system, returning null and trying to detect up and down directions was the main cause of the issue I was seeing and I don't understand how or why that is even needed?

Any thoughts?  Hopefully someone with more knowledge of the Architecture can understand the implications.  

Here is my most recent solution to the problem and examples showing the fixed behavior:

1. git clone https://github.com/cvanem/react-virtuoso/tree/scroll-element-render-fix
2. git checkout scroll-element-render-fix
3. Run the examples and look at the scroll-element-grid-scroller and scroll-element-list-scroller examples to see how the expected lists should behave.

petyosi: @cvanem The not-unmounting behavior you observe is "by design" so that the list item recalculations are minimized. The sizeRangeSystem goal is to ensure that no whitespace is present. The unmount of items no longer visible in the viewport is not a goal, but a side effect of the update. 

This behavior has been like this from the very beginning, and, as far as I am concerned, has no practical downsides - I appreciate you looking into it, but I would rather not chase that further.

cvanem: Ok, understandable.  I think we are both short on time so it is a good idea to not pursue un-necessary changes.

src/hooks/useWindowViewportRect.ts
```diff
@@ -23,39 +11,54 @@ export default function useWindowViewportRectRef(callback: (info: WindowViewport
...
-    !customScrollParent && window.addEventListener('resize', windowEH)
-    return () => {
-      element?.removeEventListener('scroll', windowEH)
-      !customScrollParent && window.removeEventListener('resize', windowEH)
+    if (customScrollParent) {
+      customScrollParent.addEventListener('scroll', scrollAndResizeEventHandler)
+      const observer = new ResizeObserver(scrollAndResizeEventHandler)
+      observer.observe(customScrollParent)
+      return () => {
+        return observer.unobserve(customScrollParent)
```
cvanem, line 51: Is this missing this in the useEffect return for the customScrollParent?

`customScrollParent.removeEventListener('scroll', scrollAndResizeEventHandler)`

cvanem: Also, the 2 examples need to be updated with the new customScrollParent prop name.

petyosi: Yes, it is!

Issue #558, opened 2022-02-06
[BUG] Tree shaking doesn't work (Table and Grid components are still included despite not importing them)


michael42:
**Describe the bug**
When importing react-virtuoso, it seems that none of the popular bundlers are able to omit unneeded parts of react-virtuoso. That makes bundle sizes larger than necessary (~24 KiB before minification/compression).

**Reproduction**
See https://github.com/michael42/react-virtuoso/tree/tree-shaking-test, I added rollup the repo and made a few test bundles, importing different things.

**To Reproduce**
1. Import only some components of react-virtuoso in a project (e.g. [here](https://github.com/michael42/react-virtuoso/tree/tree-shaking-test))
2. Bundle the result using parcel 2 or rollup (not sure about webpack)
3. Look at the produced bundle, the unused components are still included, e.g. for the linked example the output sizes are withing 15 bytes of each other (50.5 KiB)

**Expected behavior**
I expected only the `List`-Component to be included when I don't use the `Grid` or `Table`.

**Screenshots**
From a parcel report when using just `import {Virtuoso} from "react-virtuoso"`:
![image](https://user-images.githubusercontent.com/1732333/152688695-cfe11697-3cf5-4a8e-941c-f72c1904029c.png)

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: n/a

**Additional context**
I tried to look into *why exactly* tree shaking fails, but so far I had no success using `export { Table as TableVirtuoso } from './Table'` in `components.tsx` and no success manually placing `/* #__PURE__ */` annotations all over the exports and `const`s.

It's probably possible to package the library like this (but I don't think `microbundle` supports that directly):
* `index.js`: simply re-exports the other files
* `List.js`
* `Grid.js`
* `Table.js`
* `common.js`: includes the common code exported by the bundles above

A rollup-config like this produces such output:
```
export default {
    input: {
        index: 'src/index.tsx',
        List: 'src/List.tsx',
        Grid: 'src/Grid.tsx',
        Table: 'src/Table.tsx'
    },
    output: [{dir: 'build', format: 'esm'}],
    plugins: [typescript()]
}
```

What do you think, is this worth exploring/optimizing?

petyosi:
I think it might be nice to do so, although the gains are not going to be so enormous, as the list (powering the flat, grouped, and table components) is the biggest piece of code. I would love to accept a PR for that.

petyosi:
[...]

PR #580, opened 2022-02-17
feat: switch to rollup build with code-splitting

michael42:
This PR switches the build tooling away from `microbundle` (still included in `package.json` for easy comparisons) to directly using `rollup`, so the build can use rollup's code splitting. This allows bundlers to easily separate the exports of `react-virtuoso` and omit unused components. This reduces the import cost of the module by around 15% (when not using `TableVirtuoso` or `VirtuosoGrid`), and results in a roughly 8-9 KiB (minified) smaller bundle in my application, fixing #558.

I think the build is largely equivalent with a few changes:
* the `components.tsx` is split into separate files, so the code splitting works
* TypeScript is bumped a little bit to 4.4.4, otherwise an fs watcher bug keeps `rollup` running after building
* the ESM files are now named `.mjs` instead of `.m.js`, so Node.js can easily detect them
* `@babel/runtime` is used as `dependency`, because [rollup-plugin-babel recommends that for library builds](https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers)
* it uses a custom browserslist for the CommonJS build, because what [microbundle does](https://github.com/developit/microbundle/blob/master/src/lib/babel-custom.js#L54) is pretty complex


petyosi:
@michael42 this looks great, thank you very much. I will play around a bit, maybe it's worth bumping TS to the latest version.

rollup.config.js
```diff
@@ -0,0 +1,54 @@
...
+    ],
+  },
+]
+
+function babelSettings(targets) {
+  return {
+    babelrc: false,
+    babelHelpers: 'runtime',
+    extensions: ['.js', '.jsx', '.ts', '.tsx'],
+    plugins: ['@babel/plugin-transform-runtime'],
```
petyosi, line 50: From [this thread](https://github.com/preconstruct/preconstruct/issues/297) I learned that the config should contain the exact version, otherwise multiple instances of helper functions are present in the build. Perhaps there's more to be tweaked.

To see the duplicates, search for `getownproperty` in  [the build I tested](https://unpkg.com/react-virtuoso@2.8.0-alpha.1/dist/shared.mjs).

petyosi:
@michael42 I took the PR, played with it a bit, and published a couple of test versions. Bundlephobia reports the following:

[v2.7.1 - current build with microbundle](https://bundlephobia.com/package/react-virtuoso@2.7.1)
[v2.8.0-alpha.1 - a build from this PR](https://bundlephobia.com/package/react-virtuoso@2.8.0-alpha.1)
[v2.8.0-alpha.2 - a build with a fixed runtime version](https://bundlephobia.com/package/react-virtuoso@2.8.0-alpha.2)

Here's what I have noticed:
- The individual exports indeed seem smaller, but the gain is not what you report.
- The overall size has increased a bit.
- 2.8.0 starts reporting 2x size for @virtuoso.dev/urx - I'm not sure why this is the case. I still bundle the urx packages with tsdx. At some point, I tried microbundle there and it broke some features :(. 

Do you have any ideas or suggestions? I don't know how trustworthy Bundlephobia is.

michael42:
Okay, I'll have a look. I use bundlephobie myself sometimes. But I don't actually know which sizes it comares. I tried not to overdo it with complicated optimizations for this PR (that might cause a maintenance burden if it breaks anything). I *do* have a hunch that the minification by terser has something to do with it. I did most comparisons based on the non-minified output, to better see the differences. For example, the non-minified ESM rollup-build is 114.1 KiB vs 122.3 KiB (microbundle). CommonJS is much larger at 154.3 KiB vs 123.8 KiB, though.

michael42:
I did give it another shot today:

I hacked the modified rollup input/output config into microbundle to check if the size differences are really due to the different rollup settings. When building that way, microbundle outputs files with a size of 48.5 KiB (compared to the 47.1 KiB baseline). So it seems the microbundle optimizations are pretty good (and it's hard to match them). The 1.4 KiB difference is a bit larger than I'd have expected. Maybe that's because microbundle inlines the babel helpers and by splitting, some are included multiple times?

image

Then I tried (again) to re-configure microbundle so it would build that way, but unfortunately, the documented way of passing multiple files (e.g. `microbundle src/index.tsx src/Grid.tsx src/List.tsx src/Table.tsx`) causes it to generat completely separate bundles that don't reference each other. I don't think that'd be worthwhile, since that would make importing much more complicated (when using just a single component, an import like `react-virtuoso/List` has to be used, when multiple components are used, the plain `react-virtuoso` is required so no code is shared).

image

Then I tried to introduce code splitting via top-level `await import` and exporting it right away, but that errored during minify. Using `--no-compress` helped, but when building a project with parcel, the code splitting wasn't optimized away (resulting in more output files, some of which would get loaded anyway).

So I guess I'll probably give up the plain rollup build and check with the folks at microbundle, if that could support code splitting better.

Issue #497, opened 2021-11-12
[BUG] atBottomStateChange fires multiple times on mount.


JavierM42:
**Describe the bug**
The atBottomStateChange callback fires at least twice when the Virtuoso component mounts.

**Reproduction**
https://stackblitz.com/edit/react-opcwka?file=src/App.js

**To Reproduce**
Check the console and see the three logs on every reload: `false`, then `true`, then `false` again.

If you set `initialTopMostItemIndex=999999`, `atBottomStateChange` will fire a fourth time with value `true`.

If there are not enough elements to introduce a scrollbar, it will fire only two times (first `false`, then `true`).

**Expected behavior**
`atBottomStateChange` only fires once on mount, with the correct value.

**Screenshots**
<img width="404" alt="Screen Shot 2021-11-12 at 11 32 36" src="https://user-images.githubusercontent.com/32175149/141484113-d74dfd33-9715-49b3-a9ba-571d582f93bd.png">

**Desktop (please complete the following information):**
 - OS: macOS
 - Chrome

petyosi:
That would be an improvement indeed, but I am not sure when I can get to it. In the mean time, debouncing it should work as a workaround.

PR #607, opened 2022-03-11
fix: atBottomStateChange fires multiple times(#497)

BigTomM:
Fix for [bug](https://github.com/petyosi/react-virtuoso/issues/497).

Avoid an initial false/true/false flick of atBottomStateChange.

Now it only fires once(with initial state false) on mount.

_Note: The main changes are lines 192~210_

### DIFF
--- a/src/stateFlagsSystem.ts
+++ b/src/stateFlagsSystem.ts
@@ -183,17 +183,12 @@ export const stateFlagsSystem = u.system(([{ scrollContainerState, scrollTop, vi
     isAtBottom
   )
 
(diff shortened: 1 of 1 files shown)


petyosi:
I think that the change I did above does the trick. Let me know if you still experience problems with this.

hahatulk:
bug remains on version 3.1.4, fires as explained above
Debounce for 400ms doesnt help, even 1000ms

image
image

petyosi:
@hahatulk I can't debug jpegs. Use codesandbox to reproduce the problem you experience.

Issue #625, opened 2022-04-04
[BUG] validateDOMNesting(...): <tr> cannot appear as a child of <div>.


Kestutis:
**Describe the bug**
I'm using react-virtuoso with react-table and flexLayout (instead of ordinary tableLayout) and you have div`s instead of thead/tbody/tr elements. It's required to support functionality as columns sizing, etc.

Getting error `validateDOMNesting(...): <tr> cannot appear as a child of <div>.` because there are couple of components that supports only table elements as parent nodes.

```typescript
// src/components/Table.tsx
const DefaultScrollSeekPlaceholder = ({ height }: { height: number }) => (
  <tr>
    <td style={{ height }}></td>
  </tr>
)

const FillerRow = ({ height }: { height: number }) => (
  <tr>
    <td style={{ height: height, padding: 0, border: 0 }}></td>
  </tr>
)
```

**Expected behavior**
It would be nice to have an option to overwrite these components

petyosi:
@Kestutis do you want to open a PR? It should be fairly simple fix.

Kestutis:
Create PR for a fix. Also included react-table flexLayout example. There are still single issue with TableHead typings, but it's not related with flex layout.

PR #626, opened 2022-04-05
625: bugfix when using flexlayout instead of tablelayout

Kestutis:
Fixed issue when you wanted to use Table component with flex layout instead of table layout. 
https://github.com/petyosi/react-virtuoso/issues/625


petyosi:
Thank, I'll test and merge if all good

Issue #641, opened 2022-04-22
[enhancement] Add atStartThreshold


mikamations:
For prepended lists (reverse scroll) it would be great to have an atStartThreshold like atBottomThreshold so we can trigger new data API calls before hitting the top for a smoother endless scroll.

petyosi:
This is nice to have indeed. Loading new data upwards is easier though, as the top is always `0`. You can hook up to the DOM `onScroll` event or the `rangeChanged` one. 

The `atBottomThreshold` prop value controls the `followOutput` behavior, where the logic is quite complex.

PR #642, opened 2022-04-24
feat(): add atTopThreshold(#641)

BigTomM:
Add atTopThreshold, similar to atBottomThreshold

maybe help： [issue#641](https://github.com/petyosi/react-virtuoso/issues/641)

### DIFF
--- a/src/stateFlagsSystem.ts
+++ b/src/stateFlagsSystem.ts
@@ -51,12 +51,15 @@ const INITIAL_BOTTOM_STATE = {
   },
 } as AtBottomState
 
+const DEFAULT_AT_TOP_THRESHOLD = 0
+
 export const stateFlagsSystem = u.system(([{ scrollContainerState, scrollTop, viewportHeight, headerHeight, footerHeight, scrollBy }]) => {
   const isAtBottom = u.statefulStream(false)
   const isAtTop = u.statefulStream(true)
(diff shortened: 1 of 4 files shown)


src/stateFlagsSystem.ts
```diff
@@ -75,14 +78,14 @@ export const stateFlagsSystem = u.system(([{ scrollContainerState, scrollTop, vi
...
-      u.map((top) => top === 0),
+      u.combineLatest(u.duc(scrollTop), u.duc(atTopThreshold)),
+      u.map(([top, atTopThreshold]) => top <= atTopThreshold),
       u.distinctUntilChanged()
     ),
     isAtTop
   )
 
-  u.connect(isAtTop, atTopStateChange)
+  u.connect(u.pipe(isAtTop, u.throttleTime(50)), atTopStateChange)
```
petyosi, line 88: Hey, 

why did you add throttle time here? is there any scenario I'm missing?

BigTomM: image

Take a look at this example, atBottom is only true, but atTop is true-->false. I wonder if it will be a bit confusing, the two symmetrical behaviors have different trigger times, so I also added a throttle like the atBottomStateChange.

The part of the code:
image

petyosi:
Cool, I will test and merge this quite soon. Thanks!

PR #692, opened 2022-06-22
feat: add gap support to Virtuoso

mihkeleidast:
Follow-up to #686, adds gap support to regular Virtuoso list as well.

Tried to cover at least as much ground as I did with VirtuosoGrid, but considering that the regular list has lots more features, there might be some other places where gap should also be taken into account. If there are such places, let me know and I'll fix them.

### DIFF
--- a/src/hooks/useChangedChildSizes.ts
+++ b/src/hooks/useChangedChildSizes.ts
@@ -8,6 +8,7 @@ export default function useChangedListContentsSizes(
   enabled: boolean,
   scrollContainerStateCallback: (state: ScrollContainerState) => void,
   log: Log,
+  gap?: (gap: number) => void,
   customScrollParent?: HTMLElement
 ) {
   return useSizeWithElRef((el: HTMLElement) => {
@@ -30,6 +31,8 @@ export default function useChangedListContentsSizes(
       scrollHeight: (customScrollParent ?? scrollableElement).scrollHeight,
       viewportHeight: (customScrollParent ?? scrollableElement).offsetHeight,
     })
+
+    gap?.(resolveGapValue('row-gap', getComputedStyle(el).rowGap, log))
 
     if (ranges !== null) {
       callback(ranges)
@@ -74,4 +77,14 @@ function getChangedChildSizes(children: HTMLCollection, itemSize: SizeFunction, 
   }
 
   return results
+}
+
+function resolveGapValue(property: string, value: string, log: Log) {
+  if (value !== 'normal' && !value.endsWith('px')) {
+    log(`${property} was not resolved to pixel value correctly`, value, LogLevel.WARN)
+  }
+  if (value === 'normal') {
+    return 0
+  }
+  return parseInt(value, 10)
 }
--- a/src/scrollToIndexSystem.ts
+++ b/src/scrollToIndexSystem.ts
@@ -28,7 +28,7 @@ export function normalizeIndexLocation(location: IndexLocation) {
 
 export const scrollToIndexSystem = u.system(
   ([
-    { sizes, totalCount, listRefresh },
+    { sizes, totalCount, listRefresh, gap },
     { scrollingInProgress, viewportHeight, scrollTo, smoothScrollTargetReached, headerHeight, footerHeight },
     { log },
   ]) => {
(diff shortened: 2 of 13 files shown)


src/scrollToIndexSystem.ts
```diff
@@ -61,14 +61,15 @@ export const scrollToIndexSystem = u.system(
       u.pipe(
         scrollToIndex,
         u.withLatestFrom(sizes, viewportHeight, totalCount, topListHeight, headerHeight, footerHeight, log),
-        u.map(([location, sizes, viewportHeight, totalCount, topListHeight, headerHeight, footerHeight, log]) => {
+        u.withLatestFrom(gap),
```
mihkeleidast, line 64: Seems like `u.withLatestFrom` allows max 7 params, so had to add this separately. Not sure what the best pattern is in this case.

petyosi:
I'll take a look into this. This is indeed much larger and more complex.

src/hooks/useChangedChildSizes.ts
```diff
@@ -31,6 +32,8 @@ export default function useChangedListContentsSizes(
       viewportHeight: (customScrollParent ?? scrollableElement).offsetHeight,
     })
 
+    gap?.(resolveGapValue('row-gap', getComputedStyle(el).rowGap, log))
```
petyosi, line 35: Question: did you consider appending the gap value to the height of the item? It might need some logic about the last element, but the rest of the code should remain unchanged, and we won't need the gap stream.

mihkeleidast: Did not consider it really, went with a similar approach that was implemented in Grid.

I think it's better to have the gap value available separately, that can be useful in some places. Not sure how, for example, the scrollToIndexSystem would work without it. As that can scroll to an index with different alignments, we would need to know the height of the item without the gap to do center or end alignment. I think it would also be weird if the item heights would no longer be the actual item heights, but would have an arbitrary number added to all of them except the last.

petyosi:
@mihkeleidast I started testing all examples with gap "on" - https://github.com/petyosi/react-virtuoso/commit/922719e6ca1b08d08d757eb2a2f7fa9cbb5270f7. It's still in progress, but here's what I've discovered: 

http://localhost:1234/at-bottom-detection - first button + second button does not seem to work.
http://localhost:1234/decrease-items - initialTopmostItemIndex does not work as expected.
http://localhost:1234/endless-scrolling - does not work :(

The third example is the simplest one, solving that might solve the other ones. 

I also noticed that "flexing" the list makes some items "transitionally" stretching, which is not a great experience, perhaps some more flexbox voodoo can solve that. 

Thanks for the effort and for your willingness to take on that problem, it's a tough one.

mihkeleidast:
@petyosi Thanks for testing! I looked at both "at-bottom-detection" and "decrease-items", both of these should be better now - found one spot where offsets were calculated, but gap was not considered.

The "endless-scrolling" example does not seem to work in master either. Do you think it might've been broken from before?

petyosi:
> @petyosi Thanks for testing! I looked at both "at-bottom-detection" and "decrease-items", both of these should be better now - found one spot where offsets were calculated, but gap was not considered.
> 
> The "endless-scrolling" example does not seem to work in master either. Do you think it might've been broken from before?

yes, endless-scrolling is indeed broken, I am missing an e2e test there.

petyosi:
More things I saw with the same testing setup (default gap added):
http://localhost:1234/group-scroll-seek-placeholder - blinks a lot, scroller toggles. Grouping seems to blink here: http://localhost:1234/group-scroll-into-view

http://localhost:1234/header-footer - also blinking, even blocking at some point. 
http://localhost:1234/prepend-as-you-scroll - scrolling up to load new items fails to persist position.

The rest of the examples seem to be working fine.

mihkeleidast:
> More things I saw with the same testing setup (default gap added):
http://localhost:1234/group-scroll-seek-placeholder - blinks a lot, scroller toggles. Grouping seems to blink here: http://localhost:1234/group-scroll-into-view
> 
> http://localhost:1234/header-footer - also blinking, even blocking at some point.

Debugged this a bit, seems like `useChangedListContentsSizes` sends "20px" and "normal" as new gap values in an infinite loop. If I add the gap via an inline style in a custom component, it only ever sends "20px" to the stream. Feels like this is at least partially broken due to how the global testing styles are applied? Should we somehow send the gap value with a small timeout, so the styles would always be applied beforehand?

> http://localhost:1234/prepend-as-you-scroll - scrolling up to load new items fails to persist position.

I think this should be fixed now, added an additional gap calculation to `upwardScrollFixSystem` as well.

mihkeleidast:
Oh, debugged a bit more. It seems like both the top items list and the regular items list send their gap values to the same stream, which caused the infinite loop and flickering. So one way to workaround this would be to apply the test styles to both lists:
```
  <style type="text/css" id="toggle-gap">
      [data-test-id="virtuoso-item-list"],
      [data-test-id="virtuoso-top-item-list"] {
        display: flex;
        gap: 20px;
        flex-direction: column;
      }
    </style>
```

Do you think this would be reasonable, i.e. would it be expected that both lists need to have the same gap, or can we somehow use different values for those in the system?

petyosi:
Yes, it's reasonable and necessary for both top and regular containers to render content in the same way. I will update the test conditions

petyosi:
@mihkeleidast I'm happy to merge the PR in its current stage, but I do want to run one thing through you. At the moment, "gapping" the items is not so obvious - even I failed to do it right for the testing purposes, resulting in a slightly broken setup for grouping. I do want to avoid "opinionated" solutions like adding CSS classes to the list containers, but styling against the testing data attributes is not right. Direct child selectors is probably the sturdiest solution, but not one which everyone would intuitively come up with.

What do you think? How can this be made convenient?

mihkeleidast:
I personally like the approach `react-day-picker` has for this. They provide a way to override internal components, much like Virtuoso does today. In addition to that, they allow passing [`classNames`](https://github.com/gpbl/react-day-picker/blob/3bfc4eb16/packages/react-day-picker/src/types/Styles.ts#L109), which allows changing the default classnames to provide custom styles. These two approaches cover most styling solutions as well, e.g. one or both can be used with vanilla css, css modules, styled-components, etc.

In the case of Virtuoso, these do not have to have any default styles, but it could be implemented in a way that if one passes a custom classname to list_element, it will be applied to top_list_element as well. Components already work this way, I think, as the custom List component is used for both the main list and top items list.

Issue #710, opened 2022-07-13
[BUG] Gap - Cannot read property 'endsWith' of undefined


nabigraphics:
**Describe the bug**

Today, I got this error on sentry issue tracker.
I checked and it seems that undefined is coming into the value of the resolveGapValue function.

**Reproduction**

https://codesandbox.io/s/kh0fpq?file=/App.js&from-sandpack=true <- Basic Example

**To Reproduce**
1. Go to 'https://kh0fpq.csb.app/'

**Expected behavior**

**Screenshots**

<img width="1246" alt="스크린샷 2022-07-13 오후 10 10 29" src="https://user-images.githubusercontent.com/16166356/178741410-eb222627-852c-4b20-aebc-7c4aa4770284.png">

## Error log with sentry

<img width="517" alt="스크린샷 2022-07-13 오후 6 25 38" src="https://user-images.githubusercontent.com/16166356/178739617-7b8dbca0-d666-4b82-a619-48169e9b6952.png">

<img width="1335" alt="스크린샷 2022-07-13 오후 6 26 02" src="https://user-images.githubusercontent.com/16166356/178739625-f935f697-0e5e-42c3-aa55-99657dad6b92.png">

## Dev console log
<img width="569" alt="스크린샷 2022-07-13 오후 7 31 58" src="https://user-images.githubusercontent.com/16166356/178739643-e09707b4-42f8-427a-9b88-52870fb5039a.png">

**Desktop (please complete the following information):**
 - OS: Android 7
 - Browser:  Chrome Mobile WebView 64

**Additional context**

efedunin:
Hello there!
I've bumped into this issue with 2.16.6 in IE11.
It seems that method `resolveGapValue()` [here](https://github.com/petyosi/react-virtuoso/blob/v2.16.6/src/hooks/useChangedChildSizes.ts#L96) should be patched like in [this](https://github.com/petyosi/react-virtuoso/pull/711) fix.

petyosi:
@efedunin I am surprised that virtuoso works in IE11 at all. Why don't you submit a PR?

PR #711, opened 2022-07-13
fix: fix value type and add optional chaining

nabigraphics:
Fix this issue. #710


petyosi:
Thank you @nabigraphics !

PR #724, opened 2022-08-09
fix(itemSize): fix itemSize callback parameter value not as expected

simonguo:
```tsx
<Virtuoso
  itemSize={(el: HTMLElement, dimension: 'height' | 'width') => {
    // Should output `height` , but `offsetHeight`
    console.log(dimension);
  }}
/>;
```

Related:  https://github.com/petyosi/react-virtuoso/pull/688


petyosi:
@simonguo, Thanks for bringing this up.  This is entirely me being wrong and misleading the author of #688 about the signature. 

Your implementation is correct (and ergonomically better) but will open a breaking change in the actual API values passed. That's why I decided to revert the typings to reflect the actual values being passed at the moment.

PR #726, opened 2022-08-10
feat: support `align` property on `scrollIntoView`

msujew:
In my use case I want to render a list of files, where the current file should be centered only it if wasn't visible in the tree when it was selected. This didn't seem possible with the existing `scrollIntoView` API (which always scrolled the item to the bottom/top), so I added an `align` property similar to the one used for the `scrollToIndex` function.


petyosi:
That looks solid, I will look into it in the next few days.

PR #731, opened 2022-08-15
fix: fix value type and add optional chaining (IE11)

efedunin:
Fix issue https://github.com/petyosi/react-virtuoso/issues/710 for IE11


nabigraphics:
Super cool! thx @efedunin 👍

nabigraphics (approved):

PR #744, opened 2022-09-01
feat: add VirtuosoMockContext for mocking state

mihkeleidast:
Initial idea on how this could be implemented.

To use it, one would then have to mock the listState object in tests:

```tsx
<VirtuosoContext.Provider
  value={{
    listState: {
      items: data.map((i, idx) => ({
        data: i,
        index: idx,
        offset: 0,
        size: 0,
        originalIndex: idx,
      })),
    },
  }}
>
  <Virtuoso data={data} />
</VirtuosoContext.Provider>
```

However, this is not quite developer-friendly:
- it expects that you always have the mock data available in tests. In reality, the data can come from elsewhere, e.g. some mocked query result or something like that.
- having to map the data to pass forward all the different internal variables is quite cumbersome.

Ideally, I would only want to tell the context provider how many items I want to render (n or "all"). But the data is not currently available in `Items`, and not sure how to provide it there.

### DIFF
--- a/src/List.tsx
+++ b/src/List.tsx
@@ -11,6 +11,7 @@ import { listSystem } from './listSystem'
 import { positionStickyCssValue } from './utils/positionStickyCssValue'
 import useWindowViewportRectRef from './hooks/useWindowViewportRect'
 import { correctItemSize } from './utils/correctItemSize'
+import { VirtuosoContext } from './utils/context'
 import { ScrollerProps } from '.'
 
 export function identity<T>(value: T) {
@@ -153,7 +154,8 @@ const GROUP_STYLE = { position: positionStickyCssValue(), zIndex: 1, overflowAnc
 const ITEM_STYLE = { overflowAnchor: 'none' } as const
 
 export const Items = React.memo(function VirtuosoItems({ showTopList = false }: { showTopList?: boolean }) {
-  const listState = useEmitterValue('listState')
+  const ctx = React.useContext(VirtuosoContext)
+  const computedListState = useEmitterValue('listState')
 
   const sizeRanges = usePublisher('sizeRanges')
   const useWindowScroll = useEmitterValue('useWindowScroll')
@@ -169,6 +171,13 @@ export const Items = React.memo(function VirtuosoItems({ showTopList = false }: 
   const itemSize = useEmitterValue('itemSize')
   const log = useEmitterValue('log')
   const listGap = usePublisher('gap')
+
+  const listState = ctx
+    ? {
+        ...computedListState,
+        ...ctx.listState,
+      }
+    : computedListState
 
   const { callbackRef } = useChangedListContentsSizes(
     sizeRanges,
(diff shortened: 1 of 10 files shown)


petyosi:
Based on our previous conversation and the example you gave, I was thinking more about mocking the low-level missing information. Something like:

```tsx
<VirtuosoMockDomMeasures viewportHeight={700} itemHeight={700}> ... </VirtuosoMockDomMeasures>
```
(with a context provider under the hood)
From there, the hooks who attach/detach resize observers would detect a non-null value in the context and push the state. What do you think?

mihkeleidast:
@petyosi I pushed another version, instead mocking only the viewportHeight and itemHeight. In my tests, this solution works great! I assume this needs to be added to some other places as well, like the WindowViewport component, and then maybe this logic can be extracted to a separate hook, but overall, does this look like it would work for you?

src/List.tsx
```diff
@@ -395,9 +396,18 @@ export function buildWindowScroller({ usePublisher, useEmitter, useEmitterValue
...
 const Viewport: FC<PropsWithChildren<unknown>> = ({ children }) => {
+  const ctx = useContext(VirtuosoMockContext)
   const viewportHeight = usePublisher('viewportHeight')
+  const fixedItemHeight = usePublisher('fixedItemHeight')
   const viewportRef = useSize(u.compose(viewportHeight, (el) => correctItemSize(el, 'height')))
 
+  React.useEffect(() => {
+    if (ctx) {
+      viewportHeight(ctx.viewportHeight)
+      fixedItemHeight(ctx.itemHeight)
```
mihkeleidast, line 407: This seemed like the easiest thing to to here, as it does bypass the other item measuring logic, as the prop says. Not sure if there is another way of doing it, however, perhaps I did not find the correct place where the item heights are actually measured.

petyosi:
Yes, that's what I had in mind essentially. 

I think you can publish into the `sizeRanges` rather than the `fixedItemHeight`, as this property is kinda niche (and I tend to forget about it)?

mihkeleidast:
> I think you can publish into the sizeRanges rather than the fixedItemHeight, as this property is kinda niche (and I tend to forget about it)?

I did try to do that in `useChangedListContentsSizes`, but the callback returned from that hook was never called in my tests, so I couldn't mock the returned item height from there. I did not see any immediate reason for why it wasn't called...

petyosi:
ok, not a big deal, we can move forward with `fixedItemHeight`. As long as this code is test-only, it's not problematic.

mihkeleidast:
I added the publishing to List's WindowViewport component as well, so the items would be rendered when using either `useWindowScroll` or `customScrollParent`. This did not seem to be necessary for Table, I don't really know why, I would have assumed that part of the logic behaves similarly.

I also added a small docs page for this - let me know if something should be written out more clearly.

Should I add some sort of tests for this in this repo as well? There are no tests that are currently using testing-library or jsdom, so figured I'll ask before adding deps.

Anything else that should be considered here? I assume something similar needs to be added to Grid as well, but I don't have a proper implementation that uses Grid yet, so I cannot verify if it works or not. I suppose that could also be added in another PR, once necessary?

petyosi:
Thank you, I was about to suggest including a test with testing-library and JSDom + some docs. I will take a look as soon as possible.

mihkeleidast:
Added some tests! In here, the changes to TableVirtuoso's WindowViewport were necessary after all, so added the publish hooks there as well.

PR #751, opened 2022-09-13
feat: add fixed footer content to TableVirtuoso

### LINKED ISSUE #570: [Feature Request] Footer for TableVirtuoso
Could we get an option for defining a Footer component within the TableVirtuoso? 

Similar to the Footer that's available on the list, just the same thing for the table
https://virtuoso.dev/footer/

genepaul:
Closes #570 

This PR adds the `fixedFooterContent` prop to the `TableVirtuoso`. It works almost identically to `fixedHeaderContent`, except it renders a `tfoot` instead of a `thead`.

I wasn't exactly sure if there were tests I needed to add. I mostly mimicked what I saw being done with `fixedHeaderContent`. I'm pretty new to the urx framework, so forgive me if I made mistakes.

### DIFF
--- a/src/scrollToIndexSystem.ts
+++ b/src/scrollToIndexSystem.ts
@@ -29,7 +29,16 @@ export function normalizeIndexLocation(location: IndexLocation) {
 export const scrollToIndexSystem = u.system(
   ([
     { sizes, totalCount, listRefresh, gap },
-    { scrollingInProgress, viewportHeight, scrollTo, smoothScrollTargetReached, headerHeight, footerHeight, fixedHeaderHeight },
+    {
+      scrollingInProgress,
+      viewportHeight,
+      scrollTo,
+      smoothScrollTargetReached,
+      headerHeight,
+      footerHeight,
+      fixedHeaderHeight,
+      fixedFooterHeight,
+    },
     { log },
   ]) => {
     const scrollToIndex = u.stream<IndexLocation>()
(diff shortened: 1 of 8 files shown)


genepaul:
Also, sorry about the formatting changes in the files, my editor made the changes on save. I can try to revert those if you'd rather I did that.

src/scrollToIndexSystem.ts
```diff
@@ -61,65 +70,72 @@ export const scrollToIndexSystem = u.system(
...
+          ]) => {
+            const normalLocation = normalizeIndexLocation(location)
+            const { align, behavior, offset } = normalLocation
+            const lastIndex = totalCount - 1
+
+            const index = originalIndexFromLocation(normalLocation, sizes, lastIndex)
+
+            let top = offsetOf(index, sizes.offsetTree, gap) + headerHeight
+            if (align === 'end') {
+              top += fixedHeaderHeight + findMaxKeyValue(sizes.sizeTree, index)[1]! - viewportHeight + fixedFooterHeight
```
petyosi, line 89: I have not tested this (you can correct me if I am wrong), but I am not sure if the footerHeight should be added here. You can add a few buttons that call `scrollToIndex` to the example to check this.

petyosi:
Formatting is fine, it follows the eslint rules. The code looks correct, I will test it further on my side during the weekend.

src/scrollToIndexSystem.ts
genepaul: I tested this using the table-markup example, and it seems it was necessary to clear the footer when it exists. I didn't test it without a footer, but I would think footer height would be 0 in that case.

genepaul:
Thanks so much! This works great.

lucasjohnson:
Works nicely thanks!

PR #766, opened 2022-10-13
feat: support data in VirtuosoGrid

### LINKED ISSUE #523: React Virtuoso Grid
How do I get the iterator value,

`itemContent={(index, data) => {
       // here data is undefined
 }}`

mihkeleidast:
Resolves #523. Not sure if worth it, the workaround from the issue is rather simple, but as we started implementing our wrapper around VirtuosoGrid, kind of a nuisance that the component API is different from regular Virtuoso.

### DIFF
--- a/src/components.tsx
+++ b/src/components.tsx
@@ -478,11 +478,16 @@ export interface TableVirtuosoProps<D, C> extends Omit<VirtuosoProps<D, C>, 'com
   react18ConcurrentRendering?: boolean
 }
 
-export interface VirtuosoGridProps<C = unknown> extends GridRootProps {
+export interface VirtuosoGridProps<D, C = unknown> extends GridRootProps {
   /**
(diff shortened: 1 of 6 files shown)


src/components.tsx
```diff
@@ -478,12 +478,17 @@ export interface TableVirtuosoProps<D, C> extends Omit<VirtuosoProps<D, C>, 'com
   react18ConcurrentRendering?: boolean
 }
 
-export interface VirtuosoGridProps<C = unknown> extends GridRootProps {
+export interface VirtuosoGridProps<D, C = unknown> extends GridRootProps {
```
mihkeleidast, line 481: Some of the type changes could be considered breaking, IDK if that is a real concern though.

petyosi:
That's great, I will try to get to it over the weekend.

PR #768, opened 2022-10-13
feat: add Header and Footer to VirtuosoGrid

### LINKED ISSUE #197: Footer with VirtuosoGrid?
Is it possible to have a footer with `VirtuosoGrid`? Which wouldn't be part of the grid but below it. If I try to add a `footer` prop to `VirtuosoGrid` I have a typing error.

mihkeleidast:
Closes #197.

We really only need Footer for showing a loading indicator, like in the issue, but as I wanted to use similar props and examples to List, added Header as well (otherwise the `headerFooterTag` prop would be a bit weird...).

Mostly copied over the implementation from List, not sure if the heights need to be added to some calculations manually as well or will the domIoSystem forward everything automatically? The example I added seemed to work okay.

### DIFF
--- a/examples/grid-header-footer.tsx
+++ b/examples/grid-header-footer.tsx
@@ -1,0 +1,62 @@
+import * as React from 'react'
+import { GridComponents, VirtuosoGrid } from '../src'
+import styled from '@emotion/styled'
+
+const ItemContainer = styled.div`
+  box-sizing: border-box;
+  padding: 5px;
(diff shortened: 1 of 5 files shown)


petyosi:
Looks good, I am ready to merge that. Merged the data support, so this has a conflict now. Can you take a look?

mihkeleidast:
Resolved the conflicts!

Issue #769, opened 2022-10-16
[BUG] Top item rows out of order in sequential keyboard navigation (Accessibility)


tylerflo:
**Describe the bug**
For tables rendered using the `<Virtuoso>` component, the top rows (defined using the [`topItemCount` prop](https://virtuoso.dev/top-items/)) are out of order during sequential keyboard navigation (i.e. tabbing).

For example, tabbing into the table shown below does **not** select the checkbox in the top row (as would be expected from the visual order), but instead selects the checkbox in the second row (the first non-top row).

![Screen Shot 2022-10-16 at 4 42 33 PM](https://user-images.githubusercontent.com/104657426/196059818-576c1d4c-6067-4168-a999-6a7093d1aa67.png)

This is because the top rows are rendered _below_ the main content of the table in the generated HTML, which causes the web browser to think the top rows appear _after_ the main tabbable content of the table. 

This behavior causes accessibility issues. From **W3C Accessibility Guidelines (WCAG) [SC 2.4.3 Focus Order](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html)**:

> ...when users navigate sequentially through content, they encounter information in an order that is consistent with the meaning of the content and can be operated from the keyboard.

**Reproduction**
https://codesandbox.io/embed/wonderful-bogdan-oc50bh

**To Reproduce**
1. Create a table with focusable elements (including in the top rows).
2. Press Tab.

**Expected behavior**
The first focusable element in the first row of the table should be selected, regardless of whether the first row is a top row. Instead, the first focusable element in the first **non-top row** is selected.

**Desktop (please complete the following information):**
 - OS: MacOS Catalina
 - Browser: Chrome

_Note:_ Tinkering with the [`tabindex` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) may be a possible workaround, but so far I haven't been able to get it work. MDN recommends against using positive `tabindex` values anyway, since it would break the flow of other focusable elements in the page.

petyosi:
That's a valid problem. I remember putting the top row at the bottom to simplify scroll calculations and avoid z-index overrides. I am not sure when I will be able to get to trying to change it, though. @tylerflo - are you interested in a PR? FWIW, [here's the relevant part](https://github.com/petyosi/react-virtuoso/blob/master/src/List.tsx#L460-L464). I think moving it to the top should cause some breakage.

PR #803, opened 2022-12-11
fix(grouped list): #769 Move sticky container above list items contai…

jmca:
- Addresses #769 to correct natural tab ordering for accessibility
- Adds zIndex of 1 to group container
- Adjust related tests to point to the correct group element now that the DOM nodes are essentially swapped within the scroller

### DIFF
--- a/e2e/grouped-topmost-item.test.ts
+++ b/e2e/grouped-topmost-item.test.ts
@@ -10,10 +10,26 @@ test.describe('jagged grouped list', () => {
   test('puts the specified item below the group', async ({ page }) => {
     // we pick the second item, the first should remain under the group header
     const stickyItemIndex = await page.evaluate(() => {
-      const stickyItem = document.querySelector('#test-root > div > div:first-child > div > div:nth-child(2)') as HTMLDivElement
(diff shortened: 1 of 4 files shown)


petyosi:
I'll take a look around this, a bit nervous around the z-index introduction.

jmca:
@petyosi Thanks. The company I work for is in process of fulfilling AA ADA contracts, hence this issue popped up.

A possible alternative might be just to keep the group item with the rest of the scrollable items (no separate "top item" container), then zIndex _shouldn't_ be needed since it would be acting as a basic list with sticky headers. But I'm not sure if that would throw off any virtualized calculations.

petyosi:
@jmca such implementation is indeed possible, and I have that as a draft, but it requires quite a restructuring. I will look at the PR, and if there's something wrong, we can restructure this to be an optional behavior (so that we can avoid a breaking change of some sort).

jmca:
@petyosi I like the opt-in prop idea, it nails the current tab ordering case while keeping backward compatibility.

petyosi:
@jmca I made the change in v4, just published.

jmca:
@petyosi Awesome, thanks!

PR #810, opened 2022-12-22
feat: expose item data for TableRow & Item

mihkeleidast:
We're in multiple places adding `onClick` handlers for the entire table row, for example doing something like this:

```
const components = React.useMemo(
  (): TableComponents<Data> => ({
    Row: (props) => {
      const rowId = props.context?.data[props['data-item-index']]?.id;

      return (
        <TableRow
          onClick={rowId !== undefined && onRowClick !== undefined ? () => onRowClick?.(rowId) : undefined}
          {...props}
        />
      );
    },
  }),
  [onRowClick],
);
```
This means a need for accessing the row data on the TableRow (or Item for regular Virtuoso list) level. Currently we've implemented this so that we're passing the entire data array to context as well, and then accessing it via the index lookup. If the context depends on the data, then it will cause rerenders of basically everything when the data (and context) changes. 

So I looked into whether we could provide the item data to the specific components directly. It seems the data is available in the necessary places - but I'm not entirely sure if there is a good reason why it hasn't been added before. I'm also not sure how much it will help with the rerenders, but it should at least help us clean up our code a bit, as we're only passing the data to various places for this specific need.

Looking for initial feedback on this for now - do you think changing stuff like this would be viable? I expect this needs a couple of changes in other places as well, like Grid for example could benefit from the same thing, but I will add any other necessary changes once we have a general agreement.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -19,7 +19,8 @@ export interface GroupContent {
   (index: number): ReactNode
 }
 
-export type ItemProps = Pick<ComponentPropsWithRef<'div'>, 'style' | 'children'> & {
+export type ItemProps<D> = Pick<ComponentPropsWithRef<'div'>, 'style' | 'children'> & {
+  item: D
   'data-index': number
   'data-item-index': number
   'data-item-group-index'?: number
@@ -86,7 +87,7 @@ export interface GridScrollSeekPlaceholderProps {
 /**
  * Customize the Virtuoso rendering by passing a set of custom components.
  */
-export interface Components<Context = unknown> {
+export interface Components<Data = unknown, Context = unknown> {
   /**
    * Set to render a component at the top of the list.
    *
@@ -100,7 +101,7 @@ export interface Components<Context = unknown> {
   /**
    * Set to customize the item wrapping element. Use only if you would like to render list from elements different than a `div`.
    */
-  Item?: ComponentType<ItemProps & { context?: Context }>
+  Item?: ComponentType<ItemProps<Data> & { context?: Context }>
   /**
    * Set to customize the group item wrapping element. Use only if you would like to render list from elements different than a `div`.
    */
@@ -137,7 +138,7 @@ export interface Components<Context = unknown> {
 /**
  * Customize the TableVirtuoso rendering by passing a set of custom components.
  */
-export interface TableComponents<Context = unknown> {
+export interface TableComponents<Data = unknown, Context = unknown> {
   /**
    * Set to customize the wrapping `table` element.
    *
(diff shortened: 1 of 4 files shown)


src/interfaces.ts
```diff
@@ -86,7 +87,7 @@ export interface GridScrollSeekPlaceholderProps {
 /**
  * Customize the Virtuoso rendering by passing a set of custom components.
  */
-export interface Components<Context = unknown> {
+export interface Components<Data = unknown, Context = unknown> {
```
henrinormak, line 90: I guess this would be a breaking change, no? Perhaps re-order it so that Data is the second type in the generic?

src/interfaces.ts
```diff
@@ -19,7 +19,8 @@ export interface GroupContent {
   (index: number): ReactNode
 }
 
-export type ItemProps = Pick<ComponentPropsWithRef<'div'>, 'style' | 'children'> & {
+export type ItemProps<D> = Pick<ComponentPropsWithRef<'div'>, 'style' | 'children'> & {
```
henrinormak, line 22: Same sort of breaking change question here, probably should offer a default for the D type?

src/interfaces.ts
mihkeleidast: Indeed. Other interfaces that have both Data and Context always have data first, context second, so ordering it differently here would create another sort of confusion. So there is an option between consistency and a breaking change, not sure what is valued here more.

petyosi:
So my initial idea with those injectable wrapper components was to have them somewhat unaware of their content, call it separation of concerns if you wish. 

This works well in the list but imposes some practical limitations (like the one you're facing). Passing more data into the wrapper components (to me, subjectively) makes it non-obvious to the integrator if they should use the content callback prop or the item component prop because they can achieve almost the same thing in two different ways. 

The generics breaking change is also a concern, of course. We can't publish this without a major bump.

Up to what you're doing with the context (because, well, that's why this prop was introduced) - I doubt the performance impact is so large, but I might be missing something. 

Long story short - there's no technical limitation to what you're proposing (and no hidden catch) apart from the complexity and reducing the obviousness of the API.

mihkeleidast:
Thanks for the response. I agree that the current separation makes more sense for the List, but not so much for the Table, as there are HTML structure limitations with tables.

I'm not worried about the performance as such. The issue we ran into was that in some cases we had some components with state put into the table header cells, Now, when the data changed in the background, this caused all of the internal components to be redefined and rerendered, resetting the state of those subcomponents in the header cells.

Essentially, there are some subcomponents that can care about the data, and some that probably won't. Passing it through the context creates a dependency for all of the subcomponents. 

We can work around this on our side, i.e. we can adjust the definition of our component overrides so that the header row / cells do not receive the data, and as such, won't be redefined and rerendered. Or we can control the subcomponents' state externally, so it won't be reset.

We'll do the workaround on our side for now, but I'd leave this PR open for the future - in case there are any other potentially breaking changes being made, maybe this can then also be finished and included.

petyosi:
@mihkeleidast I am cooking a v4 release in [this branch](https://github.com/petyosi/react-virtuoso/tree/xmas-sprint), so I included the change.

petyosi:
@mihkeleidast @henrinormak I incorporated your changes in v4, just published.

Issue #609, opened 2022-03-15
[BUG] Vertical scroll doesn't work properly if horizontal scroll is present and moved


maciejregula:
Hello,

In the TableVirtuoso component, vertical scroll doesn't work properly if horizontal scroll exists and is moved to the right.
The bug is present even on the demo page.

**To Reproduce**
1. Go to the https://virtuoso.dev/table-fixed-columns/
2. Move the horizontal scrollbar to the right.
3. Move vertical scrollbar to the bottom - you will notice that scroll doesn't work properly

**Expected behavior**
Vertical scrollbar should properly scroll according to the current mouse position.

petyosi:
What browser/OS? Can you record a video?

maciejregula:
https://user-images.githubusercontent.com/72083355/158436494-5e194514-dc85-42b1-86e7-c0c6263de2ba.mp4

My OS: Chrome 99, windows 10 pro

fatton139:
Ran into this issue as well on Chrome.

I had a look at the call stack for when the scroll event gets is triggered on the virtuoso-scroller and it traces back to the event handler in `useScrollTop.ts`.

Not sure why the event is continuously fired.

shinenic:
Got this issue as well on my Chrome as well, but it works on my Safari.

https://user-images.githubusercontent.com/20126676/167167659-e7b996c1-5401-4ba1-b6e8-3b171bd2a1f4.mov

---

Browser: Chrome Version 100.0.4896.127 (Official Build) (arm64)
Mac OS: 12.3.1

Tirzono:
I noticed this becomes an issue when the first column is not in view anymore. So with the sticky column example it only becomes a problem if you scroll horizontally that far that even if the column would not be sticky, it would be out of view.

Maybe it is something that looks at the first column and because it's out of view it becomes an issue?

Tirzono:
After debugging for a while I noticed that it isn't necessarily an issue with the first column being the issue but it has more to do with the fact that the height and widths of the rows/columns are changing while scrolling and the reason why we saw this issue when we scrolled that far that the first column was not visible is because of the fact that the first column in the examples had a fixed width and the others had a dynamic width.

In an example where all the columns have a fixed width scrolling works normally:

https://codesandbox.io/s/sandpack-project-forked-4lf9o5?file=/App.js

So a workaround for now would be to set a fixed width for each column.

petyosi:
[...]

PR #828, opened 2023-01-23
fix: provide context to filler row

mihkeleidast:
Passes context to FillerRow, which [the props interface says the component should already receive](https://github.com/petyosi/react-virtuoso/blob/master/src/interfaces.ts#L188).

We want to use it to work around https://github.com/petyosi/react-virtuoso/issues/609, as we know the column count via the passed context.


petyosi:
Thanks! This totally makes sense.

PR #830, opened 2023-01-24
fix: VirtuosoGrid and TableVirtuoso type definiton

mihkeleidast:
Fixes multiple regressions from 397cf169c0b0447cbee875752694b89f9d235711. 

1. As `VirtuosoGridProps` accepts two generic parameters, so should the component.
2. Also, context prop was removed from `VirtuosoGridProps` the props for some reason, which I added back.
3. Also, `fixedFooterContent` prop on TableVirtuoso seems to have gone missing from the props interface, even though the implementation remained. Added this back as well, we use it.


petyosi:
I should have tagged you at that commit, knew I would miss something. Will review and merge asap

Issue #814, opened 2023-01-04
[BUG] Documentation site appears broken


six-transit:
**Describe the bug**

None of the codesandbox examples appear to be working at the moment:

![image](https://user-images.githubusercontent.com/96151753/210463211-031fe0df-ed2d-4276-b5d8-2e0e02426a09.png)

**To Reproduce**

1. Navigate to https://virtuoso.dev/
2. Scroll down a bit

**Expected behavior**

Examples on documentation site are functional

petyosi:
Thanks for that! Apparently, codesandbox is picky about `* as React` imports and modules, might be an issue in other environments too.

PR #831, opened 2023-01-24
fix: use default export from react and react-dom, build with classic jsx runtime

mihkeleidast:
Relates to https://github.com/petyosi/react-virtuoso/issues/814 and 2103a3354697081af597b43b916eae8bba415271.

The mjs extension specifically is problematic, as some bundlers expect everything from that file down to also be proper ESM. In this lib, the React imports specifically are problematic for webpack, as react is not distributed as ESM, so I get these errors after upgrading:

> ERROR in ./node_modules/react-virtuoso/dist/index.mjs 363:31-44 Can't import the named export 'createContext' from non EcmaScript module (only default export is available)

I'm not sure if changing the extension is the only solution, but it's definitely the easy one, I verified that changing this in node_modules works and requires no major changes on the consumer side (I still have to pass it through babel-loader, but that's a small thing).

I also removed the cjs extension, though that won't likely cause issues, as cjs is a bit better in terms of cross compatibility. I can revert that part, if necessary.

Figured filing a PR is easier than an issue, as the change is so small. Let me know what you think.

### DIFF
--- a/examples/auto-prepend-items.tsx
+++ b/examples/auto-prepend-items.tsx
@@ -1,5 +1,4 @@
-import { useCallback, useState } from 'react'
-import * as React from 'react'
+import React from 'react'
 import { Virtuoso } from '../src'
 import { faker } from '@faker-js/faker'
 
@@ -40,10 +39,10 @@ export function Example() {
   const START_INDEX = 10000
   const INITIAL_ITEM_COUNT = 20
 
-  const [firstItemIndex, setFirstItemIndex] = useState(START_INDEX)
-  const [users, setUsers] = useState(() => generateUsers(INITIAL_ITEM_COUNT, START_INDEX))
+  const [firstItemIndex, setFirstItemIndex] = React.useState(START_INDEX)
+  const [users, setUsers] = React.useState(() => generateUsers(INITIAL_ITEM_COUNT, START_INDEX))
 
-  const prependItems = useCallback(() => {
+  const prependItems = React.useCallback(() => {
     const usersToPrepend = 20
     const nextFirstItemIndex = firstItemIndex - usersToPrepend
 
--- a/ladle.vite.config.ts
+++ b/ladle.vite.config.ts
@@ -1,0 +1,13 @@
+/// <reference types="vitest" />
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [react()],
+  build: {
+    rollupOptions: {
+      external: ['react', 'react-dom', 'react/jsx-runtime', '@virtuoso.dev/urx', '@virtuoso.dev/react-urx'],
+    },
+  },
+})
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "e2e": "playwright test",
     "lint": "eslint src test e2e --ext .ts,.tsx",
     "typecheck": "tsc --noEmit -p tsconfig.typecheck.json",
-    "dev": "ladle serve",
+    "dev": "ladle serve --viteConfig ladle.vite.config.ts",
     "prepare": "pnpm run build && husky install",
     "semantic-release": "semantic-release"
   },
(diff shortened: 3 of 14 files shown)


petyosi:
I will have to double-check this (and, maybe, find an alternative approach) since I went back and forth on those extensions. Here's what I know and what I found out so far. 

In v4, I decided to make tree shaking work. From [what I gathered](https://dev.to/orabazu/how-to-bundle-a-tree-shakable-typescript-library-with-tsup-and-publish-with-npm-3c46), this needs `type: "module"` addition in package.json. But I might be wrong here. 

According to [the node docs](https://nodejs.org/api/packages.html#determining-module-system), `type: module` should make all files (unless explicitly marked with `cjs`) to be treated as modules. 

> Files with a .js extension when the nearest parent package.json file contains a top-level ["type"](https://nodejs.org/api/packages.html#type) field with a value of "module".

Based on the above, removing the `mjs` should not have any effect on node (but it probably has for some loaders). 

After the initial v4 release, I discovered that the file extensions are actually necessary for NextJS ([commit](https://github.com/petyosi/react-virtuoso/commit/41c45b5a80a20c8c6e2a92f2dd59f48d1d246b61)). 

Shortly after, I found out that the star imports somehow got affected by the module introduction ([issue](https://github.com/petyosi/react-virtuoso/issues/814)) in the codesandbox environment. So I standardized on named imports. 

My current understanding is that the combo of `type: module` and the extensions I ship are the right ones (in theory), but I might have to adjust (to what?) the way react is referred to (as a default import?). 

Appreciate your help on that. Also, can you help me reproduce the error you report in some sort of a stock setup? Thank you.

mihkeleidast:
> In v4, I decided to make tree shaking work. From [what I gathered](https://dev.to/orabazu/how-to-bundle-a-tree-shakable-typescript-library-with-tsup-and-publish-with-npm-3c46), this needs `type: "module"` addition in package.json. But I might be wrong here.

Not necessarily, it should be enough to have ESM distributable files together with `"sideEffects": false` in package.json, most (if not all, I haven't really done any extensive testing) bundlers can work with that.

I will try to put together a reproduction, I suppose that will also show if there is an actual issue or my projects have some deps too outdated 😅

mihkeleidast:
Here's a quite minimal webpack 4 demo with babel that shows the issue I ran into: https://stackblitz.com/edit/webpack-5-react-starter-ipsnmj?file=webpack.config.js

I also tried webpack 5, this fails with a different error (babel did not seem to have any effect here): https://stackblitz.com/edit/webpack-5-react-starter-f8qdxl?file=webpack.config.js

Note I did base these reproductions off of some minimal example I found on the internet, so not sure if updating some of the other deps might resolve the issues. Either way, forcing a major webpack or react update to use the new version of react-virtuoso would be a bummer, I think it should rather continue working without tree-shaking if the ESM files can't be properly loaded.

petyosi:
Thanks, those are certainly helpful. 

The Webpack 5 fails because of React 17. This is apparently an [annoying breaking change](https://github.com/facebook/create-react-app/issues/11769) I was not aware of. I will see if I can downgrade the compilation at the cost of a few more bytes. 

The Webpack 4 seems to be a different story, I am fairly certain that prior to react-virtuoso@4 it was picking the cjs version, since the esm one in v3 uses named imports just like v4.  

I will investigate further next week.

mihkeleidast:
Thanks! Generally I'm not opposed to having to pass react-virtuoso through babel because it's ESM, but not sure what configuration exactly to use to work around that issue. Changing the extensions back seems to work for webpack 4 + babel, unfortunate that it then breaks for Next :/

mihkeleidast:
@petyosi I did a bit of additional digging as well and found that using the simple default import from react/react-dom works as well to resolve the issue. Was the issue with CodeSandbox only the star import, would the simple default import work there as well or not?

After that, the react/jsx-runtime issue surfaced with webpack 4 as well, to fix that, I pointed your vite config to use the classic jsx runtime. I assume that should also fix the webpack 5 issues?

Pushed the changes either way.

petyosi:
Wow, that's awesome! I will take that branch and test it in the codesandbox. By the time I tested, the code did not use default exports, only named and star ones.

PR #840, opened 2023-02-01
fix: skip recalc if data changes before totalCount

### LINKED ISSUE #809: [BUG]: Remove item from list cause re-render with wrong list
**Describe the bug**
Remove an item from a list causing re-render twice. First time with new list length but not necessarily with the right items, second render will happened with the right items and the right length

**Reproduction**
Here is a Codesand: https://codesandbox.io/s/sandpack-project-forked-ogddd4?file=/App.js:409-440

**To Reproduce**
1. Remove item from the middle of the list
2. Check console.log print the rendered lists. first print list wont contain the right items, second print will contain the right ones

**Expected behavior**
After remove item should render the updated list

**Screenshots**

https://drive.google.com/file/d/1Xz5FrU5rsWsjYZbUaU2U1uWfrkuk47VS/view

mihkeleidast:
Fixes #809. Skipping the recalc if totalCount is not in sync with data.length seems to fix the issue, but I'm not entirely sure if there is maybe some better way of ensuring that the updates would be batched?

### DIFF
--- a/examples/remove-items.tsx
+++ b/examples/remove-items.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react'
+import { useState } from 'react'
+import { Virtuoso } from '../src'
+
+interface User {
+  name: string
+}
+
+const users: User[] = [
+  {
+    name: 'Sheila Robel',
+  },
+  {
+    name: 'Brian Bernier',
+  },
+  {
+    name: 'Destiny Hegmann',
+  },
+  {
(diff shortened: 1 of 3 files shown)


petyosi:
That's correct. The bogus emits at certain places in the graph this is the biggest shortcoming of the reactive state management system in use. I touched up the naming a bit to make it in sync with the other flag in there.

PR #870, opened 2023-03-29
perf: Optimize useSize performance when apply fixedItemHeight

HellowVirgil:
When apply `fixedItemHeight`, there is no need to new ResizeObserver.

It cost 2ms each item in `new ResizeObserver`.

There is almost cost 300ms in TableVirtuoso with 170 items.
image

### DIFF
--- a/src/hooks/useSize.ts
+++ b/src/hooks/useSize.ts
@@ -5,11 +5,16 @@ export type CallbackRefParam = HTMLElement | null
 export function useSizeWithElRef(callback: (e: HTMLElement) => void, enabled = true) {
   const ref = React.useRef<CallbackRefParam>(null)
 
-  let callbackRef = (_el: CallbackRefParam) => {
-    void 0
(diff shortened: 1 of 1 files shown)


petyosi:
Thanks for that. While testing your PR, I realized that there's an unnecessary re-creation of the ResizeObserver, which is actually applicable to both fixed and non-fixed item size scenarios. 

Based on that, I created a new branch with this commit: https://github.com/petyosi/react-virtuoso/commit/8258081bd85e2053943924d7dccbaa24b677b32c

This limits the ResizeObserver creation to when necessary. My guess is that this should fix your case too (as well as speed up the component re-rendering in general).

Let me know what you're seeing. If I am correct, the change above should make the change in your PR unnecessary.

petyosi:
Cool, thanks for discovering this and for the follow up. v4.1.1 is published, so I am going to close here.

Issue #889, opened 2023-05-01
[BUG] `VirtuosoGrid` + CSS Grid `List` component: items not rendered


psychedelicious:
**Describe the bug**
Conditions:
- use`VirtuosoGrid`
- use a `List` component that is a CSS grid using `gridTemplateColumns` (eg `grid-template-columns: repeat(auto-fit, minmax(100px, 1fr))`
- the client's zoom setting is not 100%
- the window/container is resized to certain widths

Result:
The rendered list's `paddingBottom` is occasionally miscalculated, resulting in a number of items not rendering in the window. It seems `paddingTop` does not get miscalculated, or its error is far less severe.

**To Reproduce**
1. Go to https://codesandbox.io/s/sandpack-project-forked-5slkdt?file=/App.js
2. Increase or decrease the browser zoom, say 120% (may need to try other zoom levels)
3. Very slowly resize the window horizontally and note when `paddingBottom` value abruptly jumps and items are rendered at a significant offset to where they should be.

**Expected behavior**
Items are always rendered in the window correctly.

**Screenshots**
Using the CSS grid visualiser helps to show the issue.
https://user-images.githubusercontent.com/4822129/235429456-19a09795-f0d9-474d-bfd0-306c446cd033.mov

**Desktop (please complete the following information):**
 - OS: macOS, windows
 - Browser: FF, chrome

**Additional context**
I'm happy to try to fix this, but I'm not 100% sure where to dig into the code. Can you point me in the right direction?

petyosi:
Thanks for the report. Looking at the example, I believe that we're facing a floating number edge case, where the calculation of how many items will fit into the width of the container is incorrect. From your clip, I am certain that's happening because of the "last item missing" state.

I will try to look at this, but due to life, it might take me a while to get on top of it. I am happy to accept a PR of course. You can give this a shot by modifying one of the examples in the `exmaples/` dir to fit the case, and then poke the [`itemsPerRow` function](https://github.com/petyosi/react-virtuoso/blob/ec97e6fa97a53c825dee3605e721ffeb6042c25d/src/gridSystem.ts#L352).

psychedelicious:
PR'd a fix

PR #892, opened 2023-05-01
fix(gridSystem): fix `itemsPerRow` precision

psychedelicious:
- Fix `itemsPerRow()` calculation to handle edge cases related to element dimension precision
- Add tests with experimentally-determined edge cases that cause precision problems with the
previous calculation
- fixes [BUG] `VirtuosoGrid` + CSS Grid `List` component: items not rendered #889

`react-virtuoso` calculates element dimensions using `getBoundingClientRect()`. This value may be
imprecise when the browser view is zoomed in/out, though technically it is [more precise than eg
`clientWidth`](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth).

To see this for yourself, use `getBoundingClientWidth()` and `clientWidth` to measure the width of
a relatively sized element at various zoom levels. `clientWidth` reports the same value each time,
while `getBoundingClientRect()` may report slightly different values.

This very small difference may be magnified during the calculation done in `itemsPerRow()` and, in
certain cases, causes the resulting value to be 1 less than what it should be. The number of items
to render in the window are thus incorrect.

My crude understanding of the cause: when the browser calculates viewport and element
dimensions at non-100% zoom level, some native rounding is involved. `getBoundingClientRect()`,
which reports precise fractional values, is affected by this rounding, while `clientWidth` rounds or
floors values. This may not be entirely correct - discussion on this is rather sparse and anyways,
it's beyond me.

### DIFF
--- a/src/gridSystem.ts
+++ b/src/gridSystem.ts
@@ -350,5 +350,5 @@ function itemTop(viewport: ElementDimensions, gap: Gap, item: ElementDimensions,
 }
 
 export function itemsPerRow(viewportWidth: number, itemWidth: number, gap: number) {
-  return max(1, floor((floor(viewportWidth) + gap) / (floor(itemWidth) + gap)))
+  return max(1, floor((viewportWidth + gap) / (floor(itemWidth) + gap)))
 }
--- a/src/gridSystem.ts
+++ b/src/gridSystem.ts
@@ -349,6 +349,6 @@ function itemTop(viewport: ElementDimensions, gap: Gap, item: ElementDimensions,
   return top > 0 ? top + gap.row : top
 }
 
-function itemsPerRow(viewportWidth: number, itemWidth: number, gap: number) {
-  return max(1, floor((viewportWidth + gap) / (itemWidth + gap)))
+export function itemsPerRow(viewportWidth: number, itemWidth: number, gap: number) {
+  return max(1, floor((floor(viewportWidth) + gap) / (floor(itemWidth) + gap)))
 }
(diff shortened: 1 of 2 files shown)


psychedelicious:
Side-note: `urx` is really cool!

src/gridSystem.ts
```diff
@@ -349,6 +349,6 @@ function itemTop(viewport: ElementDimensions, gap: Gap, item: ElementDimensions,
   return top > 0 ? top + gap.row : top
 }
 
-function itemsPerRow(viewportWidth: number, itemWidth: number, gap: number) {
-  return max(1, floor((viewportWidth + gap) / (itemWidth + gap)))
+export function itemsPerRow(viewportWidth: number, itemWidth: number, gap: number) {
+  return max(1, floor((floor(viewportWidth) + gap) / (floor(itemWidth) + gap)))
```
petyosi, line 353: Question - wouldn't it be more conservative to use `ceil(itemWidth)` instead? I understand that we're basically trying to guess if the browser will fit those in there...

petyosi:
I hear you - this is most likely the problem... Can this, however, cause the opposite kind of mistake? As in, calculate that 4 items can fit the row, while in reality, only 3 fit?

src/gridSystem.ts
psychedelicious: Actually, `ceil` is less conservative, because `itemWidth` is the divisor - `ceil`ing `itemWidth` can only result in a lesser `itemsPerRow` (and indeed, using `ceil` does not fix the problem).

The `floor(viewportWidth)` is, I believe, not needed. I cannot find a case where the calculation is different without it, but removing that `floor` does make this a more conservative calculation (erring on the side of rendering too many items in the rare edges cases). I've removed that `floor` call.

psychedelicious:
I think it is possible in some very rare edge case, but during testing of a wide range of relatively sized grid configurations, I did not see it calculating more items than it should.

I assume it is better to calculate too large of a number than too small. Too large and the change in behaviour is a performance degradation. To small and the change in behaviour is a problem with data display.

Issue #893, opened 2023-05-02
[BUG] Incompatible with `exactOptionalPropertyTypes` option


Karibash:
**Describe the bug**
When the option `extractOptionalPropertyTypes` is enabled, the optional properties of Virtuoso components cannot be passed undefined.

The type must be changed to accept undefined as shown below.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/12bdd6bfdd39be5e75a3f279705ba373a344ce63/types/react/index.d.ts#L1870

**Reproduction**
https://codesandbox.io/p/sandbox/vigilant-andras-5eoozu?file=%2Fsrc%2FApp.tsx

**To Reproduce**
1. Enable the `extractOptionalPropertyTypes` option in tsconfig.json.
2. Pass undefined to optional properties of Virtuoso components.

**Expected behavior**
Allows passing undefined to optional properties of Virtuoso components.

**Screenshots**

<img width="679" alt="スクリーンショット 2023-05-02 18 46 18" src="https://user-images.githubusercontent.com/40270352/235634232-37b66622-1929-4ed4-8eb9-9408f4a4eb7c.png">

petyosi:
What you describe as a bug is an intentional design decision - the optional properties don't accept undefined. Either set them to something meaningful or don't set them at all. 

I do understand that it might be of a slight inconvenience for you, but passing undefined to a property value is an extremely complex to handle, given the stateful nature of the component. There are multiple possible ways to handle it, with forks depending on what the previous prop value was.

Karibash:
I will share a workaround for those who are having similar problems.
https://codesandbox.io/p/sandbox/sparkling-meadow-fb9cx8?file=%2Fsrc%2FApp.tsx

petyosi:
To clarify, the workaround above is one way to handle undefined (as in, ignore it). 
However setting certain property to something and *then* setting it to `undefined` might be done with different expectations. For example, as a way to revert it to the "default" value. Or, maybe, to some other, special value.

PR #894, opened 2023-05-02
fix: make undefined accepted for optional properties

Karibash:
Change the optional property of the component to union type with undefined.

Fixes: #893


petyosi:
I'm afraid that this is not going to work like this. See my comment in #893.

PR #921, opened 2023-06-09
feat: allow deeply nested groups in GroupedVirtuoso

genepaul:
This is a draft to see if it's something you would accept. I have components that need deeply nested groups (up to 4 levels deep at this time). I'm currently accomplishing this with a custom usage of Virtuoso by indenting my items based on their level in the hierarchy. It works, but it doesn't benefit from the sticky group headers. It seemed like things were mostly setup to accommodate more than one level of groups in GroupedVirtuoso, so I thought I'd try it out.

This implementation seems to work - all tests pass, and the updated grouped example renders the groups as I would expect (though they're not indented). The only missing piece in my mind is to somehow communicate to the consumer the hierarchy of groups and items. I was thinking about tracking the `level` of groups and items and passing that to the `itemContent` and `groupContent` renderers. But I wanted to get this out here to get feedback before going any further.

The way this works, is it treats an array in `groupCounts` as another group, starting a new level. So with the following values passed to `groupCounts`: `[2, [2, 3], [2, 3, [1, 2]]]`, the levels of the first few would look like this:

image

First group has 2 items then a group that has two groups with 2 and 3 items respectively, and so on.

### DIFF
--- a/examples/grouped.tsx
+++ b/examples/grouped.tsx
@@ -4,7 +4,7 @@ import { GroupedVirtuoso } from '../src'
 export function Example() {
   return (
     <GroupedVirtuoso
-      groupCounts={Array.from({ length: 20 }).fill(3) as number[]}
+      groupCounts={[2, [2, 3], [2, 3, [1, 2]]]}
       itemContent={(index) => <div style={{ height: '20px' }}>Item {index}</div>}
       groupContent={(index) => <div style={{ height: '30px', backgroundColor: 'gray' }}>Group {index}</div>}
       style={{ height: '300px' }}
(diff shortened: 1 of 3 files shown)


petyosi:
That's a very neat usage of the component! Love what you've figured out. This being said I think that there's no advantage to embedding that within the component itself, as all the API you need is already available from the outside. To be honest, I am a bit cautious of adding that as a builtin (as it can drag people towards using the component as a treeview, which I am not mentally ready for at the moment). 

To help you further, [here's an example](https://codesandbox.io/s/zihnoc?file=/App.js&utm_medium=sandpack) of how you can pass the levels to `content`/`groupContent` through the context. 

Hope this makes sense. Let me know if this is what you're after.

genepaul:
Yeah, after reflection I've realized that I am looking for a tree-like component. My solution here doesn't allow groups to have both items and other groups in them - they can have one or the other. It appears that creating a virtualized tree component is not an easy task (I'm familiar with react-vtree and react-virtualized-tree - both appear to have been abandoned as they haven't been updated in almost 3 years).

Issue #864, opened 2023-03-16
Components["List"] expects the ref to be an HTMLDivElement in Typescript [BUG]


allanRubenstein:
**Describe the bug**
When using typescript and a custom List, Typescript throws an error and says it expects an HTMLDivElement when you try to pass the ref to anything else, for example an unordered list.

**Reproduction**
https://codesandbox.io/s/virtualized-custom-list-and-item-type-issue-g3irvh?file=/App.tsx:415-433

**To Reproduce**
1. Create a component of type `Components["List"]` following the example on https://virtuoso.dev/customize-structure/
2. Try and pass the ref to an element other than a div
5. See error on the ref

**Expected behavior**
The ref is able to be passed to any html element.

**Screenshots**
![Screenshot 2023-03-16 at 4 11 52 PM](https://user-images.githubusercontent.com/33589855/225741683-7f8864cb-8d40-4139-af73-c2dbcdfb8b31.png)

petyosi:
True, do you think you can send a PR for that? Worth doing so for the Items, too.

singhAmandeep007:
@allanRubenstein 

Following fixed the types

```
type TVirtuosoListProps = {
  children: React.ReactNode;
} & React.HTMLAttributes<HTMLUListElement>

const VirtuosoList  = forwardRef<HTMLUListElement, TVirtuosoListProps>((props, ref) => {
  return (
    <ul
      aria-labelledby="locations-list-label"
      aria-describedby="locations-list-description"
      ref={ref}
      {...props}
    >
      {props.children}
    </ul>
  );
});
```

singhAmandeep007:
@allanRubenstein 

Here is a work around 

```ts

const VirtuosoList: Components["List"] = forwardRef((props, ref) => {
  return (
    <ul
      aria-labelledby="locations-list-label"
      aria-describedby="locations-list-description"
      ref={ref as React.RefObject<HTMLUListElement>}
      {...(props as React.HTMLAttributes<HTMLUListElement>)}
    >
      {props.children}
    </ul>
  );
});

```

aPinix:
> @allanRubenstein
> 
> Here is a work around
> 
> ```ts
> const VirtuosoList: Components["List"] = forwardRef((props, ref) => {
>   return (
>     <ul
>       aria-labelledby="locations-list-label"
>       aria-describedby="locations-list-description"
>       ref={ref as React.RefObject<HTMLUListElement>}
[...]

PR #922, opened 2023-06-12
Fix typescript type errors about containers and remove unnecessary pa…

siosio34:
fix typescript error.

Links to related issues are below.
https://github.com/petyosi/react-virtuoso/issues/864

### DIFF
--- a/examples/grid-gap.tsx
+++ b/examples/grid-gap.tsx
@@ -37,7 +37,7 @@ const ListContainer = styled.div`
   flex-wrap: wrap;
   row-gap: 20px;
   column-gap: 20px;
-` as GridComponents['List']
+`
 
 export function Example() {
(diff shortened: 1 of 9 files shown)


petyosi:
I believe that this change basically leaves the `ref` as unknown. It certainly fixes the problem with styled components, but it breaks the regular component construction. Check this error after removing the ref and using the following code:

```tsx
const List: GridComponents['List'] = React.forwardRef(({ children, ...props }, ref) => {
  return <div {...props} ref={ref}>{children}</div>
}
```

image

petyosi:
The problem (I think) is that react virtuoso uses `ComponentPropsWithRef`, while the styled components use `HTMLProps`. Based [on this document](https://github.com/typescript-cheatsheets/react/blob/main/docs/advanced/patterns_by_usecase.md#componentprops), I think react-virtuoso does the right thing.

Issue #361, opened 2021-05-09
[Feature Request] Offset support for ssr


TrueCarry:
Can you please add option to render `initialItemCount` not from the start of the list?

```
const items = Array.from({ length: adjustedCount }).map((_, index) => ({ index, size: 0, offset: 0, data: data[index] }))
```
I think if you use `data[initialTopMostItemIndex + index]` instead of `data[index]` it should work, but I'm not sure how to access initialTopMostItemIndex from initialItemCountSystem.

TrueCarry:
https://github.com/TrueCarry/react-virtuoso/commit/b068fa8a96b6f0de5e95519ded7d0227d1614ac1 I've tried to make it, and in simple case it seems to work, but I'm not sure if it'll break something or even work in real world. Would be great if you give me some pointers.

petyosi:
@TrueCarry feel free to revive that by submitting a PR.

Issue #763, opened 2022-10-10
[BUG] Rerender blink when using data (server side rendering)


blahbleepew:
**Describe the bug**
When you use initial items and data together it causes a re-render even though the data hasn't changed. This happens on vite-plugin-ssr, I could not reproduce with nextjs. It doesn't seem to happen when data isn't specified. specifying the height/width of the container or the items doesn't prevent it either.

**Reproduction**
https://codesandbox.io/s/charming-tu-jj09iu?file=/pages/index/index.page.jsx

**To Reproduce**
Click refresh

**Expected behavior**
It shouldn't rerender

**Screenshots**
![image](https://s7.gifyu.com/images/ezgif-3-800f2be10b.gif)

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Brave, Chromium

petyosi:
Confirming that. It's weird that it does not happen in nextjs, though.

blahbleepew:
@petyosi I could be wrong on that. I only briefly tested with nextjs, perhaps it wasn't being server-side rendered and I didn't realize it

yokunjon:
I also experience this, using Remix, it does SSR, then on hydration rerender and flashing for a moment. I don't have a reproducable sandbox but I'll try providing one when I have the time.

Katona:
I also experience it with NextJS, too.

Melnychuk-27:
has anyone solved this problem?

12evgen:
I confirm that the error is still there. @petyosi Any ideas how to fix this?

petyosi:
Everyone, please check how it works in v4.5.1. The behavior should be fixed. 

Cheers,

PR #962, opened 2023-08-05
feat: offset the `initialItemCount` with the `initialTopMostItem`

jord-nijhuis:
Currently, when using `initialItemCount`, it would always render the first `N`
items (starting at 0). With this change, it will still render N items, but start at the offset provided by `initialTopMostItem`.

This PR contains the work performed by @TrueCarry for issue #361 plus some changes based on the feedback provided there. 

There is one issue however: due to the filter for `scrolledToInitialItem`, the rendered items will always have a `visibility: hidden` (unless the offset is 0, thus recreating the old behaviour). This would mean that, while the correct section will be rendered, it will be hidden.
https://github.com/petyosi/react-virtuoso/blob/16b418b8fae443b59ad379850b9b7d2be2c69e61/src/initialTopMostItemIndexSystem.ts#L22-L30

I have added a test case to test this, but I am not quite sure how to resolve this without breaking some of the other tests (this is my first time looking at the source code for Virtuoso). As such, some feedback would be greatly appreciated.

### DIFF
--- a/src/initialItemCountSystem.ts
+++ b/src/initialItemCountSystem.ts
@@ -2,9 +2,10 @@ import * as u from './urx'
 import { listStateSystem, buildListState } from './listStateSystem'
 import { sizeSystem } from './sizeSystem'
 import { propsReadySystem } from './propsReadySystem'
+import { getInitialTopMostItemIndexNumber, initialTopMostItemIndexSystem } from './initialTopMostItemIndexSystem'
 
 export const initialItemCountSystem = u.system(
-  ([{ sizes, firstItemIndex, data, gap }, { listState }, { didMount }]) => {
+  ([{ sizes, firstItemIndex, data, gap }, { initialTopMostItemIndex }, { listState }, { didMount }]) => {
     const initialItemCount = u.statefulStream(0)
 
     u.connect(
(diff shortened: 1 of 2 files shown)


petyosi:
Question - why do you use this? if for SEO purposes, do you know if the hidden display will affect you? 

In general, I think that what you're doing brings no trouble, so I will merge it unless I discover some problem during testing. The initial item count needs improvement, the items should not be erased (causes a brief blink), but this is not directly related to what you've changed. 

Also, can you check the conflicts?

jord-nijhuis:
> 

My main intention is serving the requested content to the end-user as quickly as possible (i.e. showing the requested item pre-rendered, and loading all the other items afterwards). SEO is something I would like to focus on in the long run. I do believe that in the past `display: none` had no effect on SEO, but I think that nowadays it does at least have some effect on the rankings.

Regarding the conflicts: I may be blind, but GitHub says that "This branch has no conflicts with the base branch". Am I looking at the wrong thing?

Thank you for your swift response, and do tell if I need to change anything!

petyosi:
Hm, I don't know if this will help much with your goal given https://github.com/petyosi/react-virtuoso/issues/763. Have you seen that? You should be affected by it. 

Also, I don't know why I see this. Not the biggest deal though, can merge it anyway.

image

jord-nijhuis:
> Hm, I don't know if this will help much with your goal given #763. Have you seen that? You should be affected by it.
> 
> Also, I don't know why I see this. Not the biggest deal though, can merge it anyway.
> 
> image

Ah, I am using Next.js, so that is perhaps why I have not encountered the bug described in #763.

I have tried rebasing my branch onto master, has this resolved the conflict message for you?

petyosi:
All good with the force push! 

I think the problem should still be there, even with NextJS (several people reporting it in the comments). Probably happening too fast and you don't see it. 

Anyway, what you've done brings no harm, and will be used when I eventually solve #763. Will review this week and update you accordingly.

PR #971, opened 2023-08-29
feat(table): preserve tr:nth-child(odd/even)

lalexdotcom:
In tables CSS styling, you often use the :nth-child(odd/even) to altern background colors for each line. With virtualisation, you can loose this: the colors are switching when an item is "undisplayed"....

With this PR, I add a second FillerRow with zero height when there's already one displayed and the first displayed item original index is even...

(only for TableVirtuoso, I still have to figure out how to do the same thing in Virtuoso - seems a bit more tricky with groups 😅)

### DIFF
--- a/examples/table-markup.css
+++ b/examples/table-markup.css
@@ -1,0 +1,3 @@
+.table table tbody tr:nth-child(odd) {
+    background-color: lightgray;
+}
--- a/examples/table-markup.tsx
+++ b/examples/table-markup.tsx
@@ -1,12 +1,15 @@
 import * as React from 'react'
 import { TableVirtuoso, TableVirtuosoHandle } from '../src/'
 
+import './table-markup.css'
+
 export function Example() {
   const ref = React.useRef<TableVirtuosoHandle>(null)
   return (
     <>
       <TableVirtuoso
         ref={ref}
+        className="table"
         totalCount={1000}
         components={{
           EmptyPlaceholder: () => {
--- a/src/TableVirtuoso.tsx
+++ b/src/TableVirtuoso.tsx
@@ -1,6 +1,6 @@
 import { systemToComponent } from './react-urx'
 import * as u from './urx'
-import React from 'react'
+import React, { Fragment } from 'react'
 import useChangedListContentsSizes from './hooks/useChangedChildSizes'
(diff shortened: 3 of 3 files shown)


petyosi:
@lalexdotcom thanks for raising this. The code and your reasoning is legit. However, I am paranoid about this being a some kind breaking change with unforeseen side effects. People tend to do the darnest things with the table. 

In case you overlooked this, alternating rows can be done with custom containers - [here's an example with the table](https://codesandbox.io/s/table-even-odd-styling-9v7vvh). Same can be done for the lists, since you have access to those values in the wrapper. 

I will give this some further thought. Of course, the elegance of the pseudoselectors is superior to the solution I've presented above.

lalexdotcom:
@petyosi your point of view makes perfect sense...
You're right, there's might be side effects to this PR... and the solution you gave is quite elegant from a code point of view, and reading your comment makes me thinks about more complex situations you can handle by code but not CSS with Virtuoso even after this PR (let's say... `nth-child(3n+1)`)

But from a CSS point of view, to make a customizable component, it would take some weird and very constraining properties: (`oddBackgroundColor` / `evenBackgroundColor` in this use case)....

I actually dont' have any wide generic solution: my use case was for odd/even, so I figured out a solution compatible with CSS 😅

Which makes me wonder (this should be in discussions instead of PR comments): is virtualization the solution for too much DOM nodes or too much **_displayed_** DOM nodes? Actually, the question underneath is "Does a `display:none` or `visibilitty:hidden` HTML element use rendering performances?"... using virtualization to control only the display or visibility style attribute could then be a good solution to maintain the original DOM...

petyosi:
@lalexdotcom the approach I've taken mitigates the quirks imposed by the virtualization, as it relies on "true" numbers (as the DOM can't be trusted). 

Regarding your more general concern, here's a few great (IMO) takes on the matter: https://github.com/petyosi/react-virtuoso/discussions/959.

lalexdotcom:
I actually chose a "hybrid" method: 
```css
[data-part="row"][data-odd] {
	background-color: lightgray
}
```
```typescript
// ...
	components={{
		TableRow: props => {
			return (
				<tr
					data-part="row"
					{...{ [props['data-index'] % 2 ? 'data-odd' : 'data-even']: true }}
					{...props}
				></tr>
			);
		},
	}}
//...
```

Thanks for your kind feedback ;-)

PR #983, opened 2023-09-18
feat: added table-state

ru-van-urk:
I noticed that the table docs mentioned `restoreStateFrom` but this feature is not implemented. This pull request adds that functionality. If there is more to be done I will happily do so. I tested the functionality in `table-state.tsx`.

Also whilst I am working on tables. I would love to update the react-table integration documentation because the package is updated, nowadays it is tanstack table. My question is should I leave the react-table docs as is for backwards compatibility and create a seperate example called tanstack/table or should I replace it alltogether?

### DIFF
--- a/examples/table-state.tsx
+++ b/examples/table-state.tsx
@@ -1,0 +1,54 @@
+import * as React from 'react'
+import { StateSnapshot, TableVirtuoso, TableVirtuosoHandle } from '../src/'
+
+export function Example() {
+  const ref = React.useRef<TableVirtuosoHandle>(null)
+  const state = React.useRef<StateSnapshot | undefined>(undefined)
+  const [key, setKey] = React.useState(0)
+
+  console.log('Rendering with key', key)
+  return (
+    <div>
+      <button
+        onClick={() => {
+          ref.current?.getState((snapshot) => {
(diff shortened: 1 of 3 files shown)


petyosi:
Lovely! I will test the PR and merge it if everything goes ok. 

re: react-table example. Yes, please go ahead. I'm happy to accept a PR on the matter. I am not an user of the react-table/tanstack table, so I am not sure what people need there, but last time I tried to integrate it, it was quite an adventure.

petyosi:
About backwards compatibility - no, no need to keep that.

examples/table-state.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react'
+import { StateSnapshot, TableVirtuoso, TableVirtuosoHandle } from '../src/'
+
+export function Example() {
+  const ref = React.useRef<TableVirtuosoHandle>(null)
+  const state = React.useRef<StateSnapshot | undefined>(undefined)
+  const [key, setKey] = React.useState(0)
+
+  console.log('Rendering with key', key)
```
jibbewillem, line 9: ^

jan-martenn (approved):

PR #1061, opened 2024-04-05
fix: stop "Zero-sized element, this should not happen" error from showing up when it should not

### LINKED ISSUE #1060: [BUG] Zero-sized element, this should not happen
**Describe the bug**
The error "zero-sized element, this should not happen" appears as of upgrading to v4.7.7. There is also a react-virtuoso internal exception along the lines of "Cannot read properties of null (reading 'dataset')".

This commit seems to have introduced the issue: https://github.com/petyosi/react-virtuoso/commit/1f8c8ee833fe1efafe4879e709316d0c4812bb9c

**Reproduction**
Use virtualized lists and click around so that the underlying data changes, i.e., re-renders/mounts/unmounts occur.

Sorry, I don't have time for a codesandbox. If you don't like that, feel free to close my report. I'm locking my version to 4.7.6.

**Screenshots**
image

**Additional context**
This error has been discussed in other issues such as #433 and #206.

DovydasSp:
this should fix #1060 

Very ugly and extreme example where it happens: https://codesandbox.io/p/sandbox/red-glitter-hcqz3d?file=%2FApp.js%3A62%2C1 
After fix: https://codesandbox.io/p/sandbox/hungry-roentgen-3ff9yp

### DIFF
--- a/src/hooks/useSize.ts
+++ b/src/hooks/useSize.ts
@@ -12,12 +12,12 @@ export function useSizeWithElRef(callback: (e: HTMLElement) => void, enabled = t
   if (typeof ResizeObserver !== 'undefined') {
     const observer = React.useMemo(() => {
       return new ResizeObserver((entries: ResizeObserverEntry[]) => {
-        const element = entries[0].target as HTMLElement
-        if (element.offsetParent !== null) {
-          requestAnimationFrame(() => {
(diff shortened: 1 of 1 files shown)


petyosi:
That's the way to go. I will test and likely merge over the weekend.

PR #1073, opened 2024-04-17
fix: adjust the way used to set data-test-id for TableBody

ts1994tw:
## Overview
I'm trying to integrate `react-virtuoso` with MUI table in my current project, also it is also based on testing-library: `ByTestId`https://testing-library.com/docs/queries/bytestid/, when tried to use `ByTestId` and found it probably might be needed to change

## Changes
- instead of setting table with test id like `data-test-id`, but set table with `data-testid` as its test id

## Check
Screenshot 2024-04-17 at 14 55 41

## Conclusion
If it seems like a good fix, it'd be cool to see it become part of the library. Thanks for your time, and kudos for your fantastic works in open-source 🙇

### DIFF
--- a/src/TableVirtuoso.tsx
+++ b/src/TableVirtuoso.tsx
@@ -163,7 +163,7 @@ const Items = /*#__PURE__*/ React.memo(function VirtuosoItems() {
 
   return React.createElement(
     TableBodyComponent,
-    { ref: callbackRef, 'data-test-id': 'virtuoso-item-list', ...contextPropIfNotDomElement(TableBodyComponent, context) },
+    { ref: callbackRef, 'data-testid': 'virtuoso-item-list', ...contextPropIfNotDomElement(TableBodyComponent, context) },
     [paddingTopEl, ...items, paddingBottomEl]
   )
 })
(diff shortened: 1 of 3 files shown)


petyosi:
I'm ok with the fix, by the time I used that my understanding of the `testid` attribute was insufficient. However, there's more to be done to make the fix uniform - there are several other `data-test-id` attributes in the table and in the virtuoso component, and the e2e tests are using an approach that looks like this:

```tsx
      const stickyItem = document.querySelector('[data-test-id=virtuoso-top-item-list] > div') as HTMLElement
```
If you have the capacity, I'm happy to accept a larger PR that replaces all of the test attributes with the more conventional `data-testid`, but the e2e tests need to be updated too. 

Hope this makes sense, thank you for the effort.

ts1994tw:
@petyosi Thanks for your time on reviewing this PR.

I'm good with putting efforts on adjustments which are needed. But compared to my branch name here `boren/fix-the-way-used-to-set-data-test-id-for-table`, I would be happy to have this PR merged first and I will create another PR for any needed adjustments about `data-testid` soon.

petyosi:
Merged in https://github.com/petyosi/react-virtuoso/releases/tag/v4.7.9, due to the rebase GitHub did not pick this up.

PR #1090, opened 2024-05-21
feat: adds itemClassName/listClassName to Virtuoso

ctate:
Hey @petyosi! I have submitted this PR to add `itemClassName` and `listClassName` props to the `<Virtuoso />` component. I tried my best to make sure this change aligns with the existing patterns used in `<VirtuosoGrid />`.

I am open to feedback and willing to make any necessary adjustments to this PR. Thank you for your work on this project, and I look forward to your thoughts on these additions!


petyosi:
I don't want to pollute the props any further, given that this is already doable. The class names in the grid are necessary for the Grid layout to be created. In the list component, this is not the case. If you need to target the elements at that level you can achieve the same by passing custom components or by using good old-school class + element selectors.

PR #1109, opened 2024-07-12
feat: add memo for useSize argument

dickeylth:
Add `React.useMemo` for `useSize` input param to prevent `callback` param may accidentally updated, which makes unnecessary [ResizeObserver recreation](https://github.com/petyosi/react-virtuoso/blob/master/src/hooks/useSize.ts#L14), to speed up scroll performance.


petyosi:
Nice find!

Issue #1081, opened 2024-05-04
[BUG]  Simple header breaks restore state location


phazei:
**Describe the bug**
If restoreStateFrom is used along with components Header, then the restoreStateFrom does not restore to the correct location

**Reproduction**
https://codesandbox.io/p/sandbox/sandpack-project-forked-x869rf?file=%2FApp.js%3A193%2C15

**To Reproduce**
Scroll up, click the "save" button.  Click the "restore" button.  See the location jump.

**Expected behavior**
Comment out line 194. 
```
            components={{
                 //Header: Header,
            }}

```
Scroll up, click the "save" button.  Click the "restore" button.  See the location not jump.

phazei:
I tried to debug this issue myself for a few hours, but honestly I just don't understand the urx streams at all, I can't even manage to figure out how to follow the flow.

In the stateLoadSystem, I found this which I think is the getState callback
```
    u.subscribe(
      u.pipe(getState, u.withLatestFrom(sizes, scrollTop, useWindowScroll, statefulWindowScrollContainerState, statefulWindowViewportRect)),
      ([callback, sizes, scrollTop, useWindowScroll, windowScrollContainerState, windowViewportRect]) => {
        const ranges = sizeTreeToRanges(sizes.sizeTree)
        if (useWindowScroll && windowScrollContainerState !== null && windowViewportRect !== null) {
          scrollTop = windowScrollContainerState.scrollTop - windowViewportRect.offsetTop
        }
        callback({ ranges, scrollTop })
      }
    )

```
And it seems that when I log the scrollTop is saving the correct position.  I tried modifying the state.tsx example by adding a 200px height header.  If I am at the very top and I save, the scrollTop is 0, when I restore it seems to compensate for the header a second time and it scrolls down 200px in addition to the scrollTop value.  So if I scroll to item 0, and save, scrollTop is already 200px, and when I restore, it goes to item 8, which is 400px.  It works perfectly if I manually subtract the header height from scrollTop before I save.  Though I tried to retrieve the value from the stream it seems to be in, I see it grabbed from useEmitterValue in Virtuoso.tsx  but whatever I tried I couldn't get it.  I'd really love to help and debug it myself, but I can't even figure out where it's restoring it from.

petyosi:
This is most likely a legit issue, but I am stretched very thin with my time; it's unlikely for me to be able to get to it at any point soon.

phazei:
I think I figured out a solution.

[...]

PR #1110, opened 2024-07-14
fix: account for headerHeight when restoring state

phazei:
RE: https://github.com/petyosi/react-virtuoso/issues/1081

### DIFF
--- a/examples/state.tsx
+++ b/examples/state.tsx
@@ -1,6 +1,10 @@
 import * as React from 'react'
 import { StateSnapshot, Virtuoso, VirtuosoHandle } from '../src'
 
+function Header() {
+  return <div>Header</div>
+}
+
 export function Example() {
   const ref = React.useRef<VirtuosoHandle>(null)
   const state = React.useRef<StateSnapshot | undefined>(undefined)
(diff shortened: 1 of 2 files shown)


petyosi:
Thanks, I will test and merge if no issues.

phazei:
So, it looks like my original change was modified and moved into the stateLoadSystem.ts where it does the `scrollTop -= headerHeight`.

The issue with that is `headerHeight` isn't loaded at that point yet, it's value is always 0, so it doesn't seem to fix the issue.

In the original PR it was in the initialTopMostItemIndexSystem.ts.
```
        skipFrames(4, () => {
          u.handleNext(scrollTop, () => {
            u.publish(scrolledToInitialItem, true)
          })

          //by this frame the values have loaded if they exist
          const currentHeaderHeight = u.getValue(headerHeight)
          if (typeof initialTopMostItemIndex === 'object' && typeof initialTopMostItemIndex.offset === 'number') {
            initialTopMostItemIndex.offset -= currentHeaderHeight
          }

          u.publish(scrollToIndex, initialTopMostItemIndex)
        })
```

Those skipFrames is very important.  I did test different skipFrames, and unless it's at least 3, the value of headerHeight isn't yet available.  I don't know where or when it's calculated, but it needs to be sooner if it's going to be inside stateLoadSystem, because header isn't available as it is now.

petyosi:
@phazei - the commit I pushed includes an example. As far as I remember, it works as expected. Can you elaborate? Perhaps a sandbox would help me understand the problem better.

phazei:
I'm currently investigating, things might be pointing to something changing in our codebase, I haven't figured out why it's ignoring the height currently.  Will update when I do!

phazei:
Seems to have gone away, phantom issue 🤷   working fine now

PR #1147, opened 2024-10-13
feat: topItemCount for TableVirtuoso

### LINKED ISSUE #1130: TableVirtuoso does not render topItemCount items
**Describe the bug**
When using the topItemCount prop for the TableVirtuoso, the items are not rendered. They appear to be an empty div on top of the other items.

**Reproduction**
https://codesandbox.io/p/sandbox/sandpack-project-forked-qflzwl

yuta4j1:
Resolves #1130 .
I wanted to separate topItems from items as in `Virtuoso.tsx`, but in the case of `TableVirtsuoso.tsx`, the Items component contains a TableBodyComponent, so I could not put a TableRowComponent outside of it, and So, this is how it is implemented.
If you have any ideas for a better implementation, I'd be happy to point them out.

### DIFF
--- a/src/TableVirtuoso.tsx
+++ b/src/TableVirtuoso.tsx
@@ -79,7 +79,7 @@ const DefaultFillerRow = ({ height }: { height: number }) => (
 
 const ITEM_STYLE = { overflowAnchor: 'none' } as const
 
-const Items = /*#__PURE__*/ React.memo(function VirtuosoItems() {
+const Items = /*#__PURE__*/ React.memo(function VirtuosoItems({ showTopList = false }: { showTopList?: boolean }) {
   const listState = useEmitterValue('listState')
   const sizeRanges = usePublisher('sizeRanges')
   const useWindowScroll = useEmitterValue('useWindowScroll')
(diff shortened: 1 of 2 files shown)


src/TableVirtuoso.tsx
```diff
@@ -123,12 +123,51 @@ const Items = /*#__PURE__*/ React.memo(function VirtuosoItems() {
...
   const statefulTotalCount = useEmitterValue('statefulTotalCount')
   const context = useEmitterValue('context')
+  const fixedHeaderHeight = useEmitterValue('fixedHeaderHeight')
 
   if (statefulTotalCount === 0 && EmptyPlaceholder) {
     return <EmptyPlaceholder {...contextPropIfNotDomElement(EmptyPlaceholder, context)} />
   }
 
-  const paddingTop = listState.offsetTop + paddingTopAddition + deviation
+  const topItems = (showTopList ? listState.topItems : []).map((item) => {
```
petyosi, line 132: This is duplicating the `const items` block below. I would rather have this refactored into a common routine to reduce size and avoid potential bugs because things changed in one of the blocks but were forgotten into the other.

src/TableVirtuoso.tsx
```diff
@@ -123,12 +123,51 @@ const Items = /*#__PURE__*/ React.memo(function VirtuosoItems() {
...
+    const index = item.originalIndex!
+    const key = computeItemKey(index + firstItemIndex, item.data, context)
+    const offsetTop = listState.topItems.reduce((acc, item, itemIndex) => {
```
petyosi, line 135: This nested loop can theoretically cause performance problems.

petyosi (changes requested):

petyosi:
Looks good, I will test and merge if everything is ok.

src/TableVirtuoso.tsx
yuta4j1: I made topItems and items common. Does it match your image?
78bd0c6bba33d98b33a48325560b4de48883a436

src/TableVirtuoso.tsx
yuta4j1: Indeed. I made a list of offsets in advance and used them in the loop.
https://github.com/petyosi/react-virtuoso/pull/1147/files#diff-bbd02b2ee6e5b3189b3c2bcbed9062a411c4ad1a500e9e8fd47777a5a65d25b5R93-R100

petyosi:
@yuta4j1 Thank you for your contribution :).

Issue #1087, opened 2024-05-21
[BUG] VirtuosoGrid - endReached doesn't call if data set is less then the container


dinataranis:
**Describe the bug**
I am using `VirtuosoGrid` component with `endReached` method and if an initial data set is less then the container the `endReached` is not fired (if I switch to `Virtuoso` component `endReached` is fired)
I have checked the similar issues: 
https://github.com/petyosi/react-virtuoso/issues/919
https://github.com/petyosi/react-virtuoso/issues/924
and as I see `VirtuosoGrid` and `Virtuoso` components behave differently towards `endReached` method if the initial data set is less then the container

**Reproduction**
[Virtuoso with endReached works as expected](https://codesandbox.io/p/sandbox/virtuoso-end-reached-46qzg4)
[VirtuosoGrid with endReached the initial fire doesn't work ](https://codesandbox.io/p/sandbox/virtuoso-grid-end-reached-g2gvp4)

**Expected behavior**
I expect both `Virtuoso` and `VirtuosoGrid` to work the same way and to fire the initial `endReached` if the data set is less then the container (at least one time)

erdemdev:
I ran into this issue too. I'm glad I checked the issues section; otherwise, I would still be struggling with it. For now, I worked around it using the Intersection Observer API.

stuartambient:
i found that endReached doesn't trigger at all even if data is more than container.  There is nothing in the docs that confirms endReached is part of VirtuosoGrid.

aderiushev:
same here

prasannamestha:
+1 facing this issue. Will take a look over the weekend to see if I can raise a fix 👍

prasannamestha:
@petyosi Browsing through the code it seems like this file `gridSystem.ts`  on line 315: `u.withLatestFrom(totalCount, hasScrolled),` uses `hasScrolled` value to trigger `endReached`. 

When the container size is greater than the list size, the `hasScrolled` value is always false and usually there is no scrolling available - hence this function will never be called. Would it be a good idea to omit the `hasScrolled` check here?

prasannamestha:
Raised a PR: https://github.com/petyosi/react-virtuoso/pull/1166 - please take a look

say8425:
In my case, the issue was an over-injected `increaseViewportBy`. Consider reducing the `increaseViewportBy` value or removing it altogether.

prasannamestha:
@petyosi I believe we can close this issue now

PR #1166, opened 2024-11-30
fix(gridSystem.ts): call onEndReached when the data is less than cont…

prasannamestha:
This PR fixes this bug: https://github.com/petyosi/react-virtuoso/issues/1087

### DIFF
--- a/src/gridSystem.ts
+++ b/src/gridSystem.ts
@@ -310,23 +310,23 @@ export const gridSystem = /*#__PURE__*/ u.system(
 
     const endReached = u.streamFromEmitter(
       u.pipe(
-        u.combineLatest(gridState, data),
+        u.combineLatest(gridState, totalCount),
         u.filter(([{ items }]) => items.length > 0),
-        u.withLatestFrom(totalCount, hasScrolled, data),
-        u.filter(([[gridState, data], _totalCount, hasScrolled]) => {
+        u.withLatestFrom(hasScrolled),
+        u.filter(([[gridState, totalCount], hasScrolled]) => {
           const lastIndex = gridState.items[gridState.items.length - 1].index
-          const isLastItemRendered = lastIndex === (data?.length || 0) - 1
+          const isLastItemRendered = lastIndex === totalCount - 1
 
           // User has scrolled
           if (hasScrolled) return isLastItemRendered
 
           // User has not scrolled, so check whether grid is fully rendered
           const isFullyRendered =
-            gridState.bottom > 0 && gridState.itemHeight > 0 && gridState.offsetBottom === 0 && gridState.items.length === data?.length
+            gridState.bottom > 0 && gridState.itemHeight > 0 && gridState.offsetBottom === 0 && gridState.items.length === totalCount
 
           return isFullyRendered && isLastItemRendered
         }),
-        u.map(([_, totalCount]) => {
+        u.map(([[, totalCount]]) => {
           return totalCount - 1
         }),
         u.distinctUntilChanged()
--- a/src/gridSystem.ts
+++ b/src/gridSystem.ts
@@ -253,17 +253,21 @@ export const gridSystem = /*#__PURE__*/ u.system(
             } else {
               startIndex = perRow * floor((startOffset + rowGap) / (itemHeight + rowGap))
               endIndex = perRow * ceil((endOffset + rowGap) / (itemHeight + rowGap)) - 1
-              endIndex = min(totalCount - 1, max(endIndex, perRow - 1))
+              endIndex = min((data?.length || totalCount || 0) - 1, max(endIndex, perRow - 1))
               startIndex = min(endIndex, max(0, startIndex))
             }
 
             const items = buildItems(startIndex, endIndex, data)
             const { top, bottom } = gridLayout(viewport, gap, item, items)
-            const rowCount = ceil(totalCount / perRow)
-            const totalHeight = rowCount * itemHeight + (rowCount - 1) * rowGap
-            const offsetBottom = totalHeight - bottom
+            // const rowCount = ceil(totalCount / perRow) // this is the total number of rows based on `totalCount`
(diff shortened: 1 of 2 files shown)


src/gridSystem.ts
```diff
@@ -253,17 +253,21 @@ export const gridSystem = /*#__PURE__*/ u.system(
             } else {
               startIndex = perRow * floor((startOffset + rowGap) / (itemHeight + rowGap))
               endIndex = perRow * ceil((endOffset + rowGap) / (itemHeight + rowGap)) - 1
-              endIndex = min(totalCount - 1, max(endIndex, perRow - 1))
+              endIndex = min((data?.length || totalCount || 0) - 1, max(endIndex, perRow - 1))
```
petyosi, line 256: Why is this change necessary?

src/gridSystem.ts
```diff
@@ -272,9 +276,12 @@ export const gridSystem = /*#__PURE__*/ u.system(
...
       u.pipe(
-        data,
-        u.filter((data) => data !== null),
-        u.map((data) => data!.length)
+        u.combineLatest(data, totalCount),
+        u.filter(([data]) => data !== null),
+        u.map(([data, totalCount]) => {
+          return Math.max(data!.length, totalCount)
+        }),
+        u.distinctUntilChanged()
```
petyosi, line 284: Why are you changing this? The component should accept either data or total count, and if data is supplied, it should control the total count value throughout the rest of the implementation.

src/gridSystem.ts
```diff
@@ -253,17 +253,21 @@ export const gridSystem = /*#__PURE__*/ u.system(
...
+            // const rowCount = ceil(totalCount / perRow) // this is the total number of rows based on `totalCount`
+            const rowsVisibleCount = ceil((data?.length || totalCount || 0) / perRow) // number of rows rendered on screen
 
-            return { items, offsetTop: top, offsetBottom, top, bottom, itemHeight, itemWidth } as GridState
+            // const totalHeight = rowCount * itemHeight + (rowCount - 1) * rowGap // this is the total height that the container should occupy
+            const virtualHeight = rowsVisibleCount * itemHeight + (rowsVisibleCount - 1) * rowGap // height visible
+
+            const offsetBottom = virtualHeight - bottom
+
+            return { items, offsetTop: top, offsetBottom, top, bottom, itemHeight, itemWidth, data } as GridState
```
petyosi, line 270: Unless I am missing something, this change should not do anything. See my comment on the relationship between totalCount and data.

src/gridSystem.ts
```diff
@@ -310,11 +317,23 @@ export const gridSystem = /*#__PURE__*/ u.system(
...
+            (gridState.bottom !== 0 &&
+              gridState.itemHeight !== 0 &&
+              gridState.offsetBottom === 0 &&
+              gridState.items.length === data?.length &&
+              // gridState.items[gridState.items.length - 1].index < totalCount - 1
+              gridState.items[gridState.items.length - 1].index === (data?.length || 0) - 1)
+          )
+        }),
+        u.map(([[_, data]]) => {
+          return data?.length
```
petyosi, line 335: How does that work if the grid has no data and uses totalCount?

src/gridSystem.ts
prasannamestha: I'll prepare a video later today to explain but here's the gist:

When both `data` and `totalCount` is passed to `VirtuosoGrid` - the endIndex calculation was faulty. Now looking at this after a good night sleep, it seems like we might not require this as we update `totalCount` below. I'll get back on this

src/gridSystem.ts
prasannamestha: My bad, I added `data` here for testing, forgot removing it. Will update this later today

src/gridSystem.ts
prasannamestha: But it is a common use case for something like a dynamic list (or infinite list)

src/gridSystem.ts
prasannamestha: Okay here's the explanation why this change is necessary:

In a component with paginated data (data received from `useInfiniteQuery`), `onEndReached` function fires the callback to load more data - that eventually increases the data size. Additionally there might be a different `totalIndex` specified to calculate the container height. So usually `totalCount` could be greater than or equal to `data.length`

Hence the `endIndex` should consider the min of `data.length` or `totalCount` to get the right value

src/gridSystem.ts
petyosi: I can see how one can potentially achieve some elaborate implementation using a combination of both props and load on demand, and conditional rendering depending on data being present or not.

However, this is way beyond trivial, and supporting this would need a good set of examples and tests. If these are not present, the capability would just result in people struggling infinitely and claiming "bugs" that I need to deal with.

prasannamestha: I see.. Gotcha, maybe I touched a bit more code than necessary. I'll make some changes

prasannamestha: @petyosi take a look now, I have added an example page and updated code to only touch `endReached` fn.

src/gridSystem.ts
```diff
@@ -310,11 +310,25 @@ export const gridSystem = /*#__PURE__*/ u.system(
...
+        u.filter(([[gridState, data], _totalCount, hasScrolled]) => {
+          const lastIndex = gridState.items[gridState.items.length - 1].index
+          const isLastItemRendered = lastIndex === (data?.length || 0) - 1
+
+          // User has scrolled
+          if (hasScrolled) return isLastItemRendered
+
+          // User has not scrolled, so check whether grid is fully rendered
+          const isFullyRendered =
+            gridState.bottom > 0 && gridState.itemHeight > 0 && gridState.offsetBottom === 0 && gridState.items.length === data?.length
```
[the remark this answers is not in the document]
prasannamestha: It's strange but using `totalCount` here doesn't fire more than once. But I think I found a way to get around it. Hold on, sending an update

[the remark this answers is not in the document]
prasannamestha: @petyosi check now :)

PR #1276, opened 2025-08-08
feat: Scroll on refresh

langleyd:
For improved accessibility it can be nice to scroll lists to keep the selected item in the viewport when data changes(inserted, filters applied etc).

Having not been able to get all the way there with the existing apis, I've added a new one for your consideration. It adds a `targetsNextRefresh` param to `scrollIntoView`, that defers the scroll till after the next data update and before subsequent repaints. Thanks!

Existing Apis tried first:
- `initialTopMostItemIndex` only works for the first render. 
- Combining this with `key` means unmounting/remounting the component when changing the data which can be quite slow for the user(E.g. just trying to filter the data)
- Using `scrollIntoView` in useEffect with a timeout works but it renders the top of the list first and the scrolls down which is looks a bit glitchy

### DIFF
--- a/apps/virtuoso.dev/docs/guides/virtuoso/scroll-on-refresh.md
+++ b/apps/virtuoso.dev/docs/guides/virtuoso/scroll-on-refresh.md
@@ -8,11 +8,11 @@ sidebar_position: 121
 
 ## Scrolling an item into view when the data changes
 
-When data in a list changes(filters being applied or new data being inserted) important information may be pushed off the screen(like a selected item).
+When data in a list changes (E.g. filters being applied or new data being inserted) important information may be pushed off the screen(like a selected item).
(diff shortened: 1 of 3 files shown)


petyosi:
Thanks for your effort. I understand where this is coming from, and this is one of the reasons behind the [imperative data API design of the message list](https://virtuoso.dev/virtuoso-message-list/working-with-data/#replacing-data). 

Back to your point, changing the key is not the worst performance crime I believe. If you're filtering, the data is going to be "entirely new", so the existing size cache is going to be totally wrong (and, in fact, hinder performance). Most of the HTML will be remounted anyway, because of the data change.

petyosi:
@langleyd thanks for the clarification, I understand the reasoning. Let me think this through over the weekend, I really appreciate your effort of digging into the arcane aspects of my code.

petyosi:
I have an alternative proposal in #1277. Please check.

PR #1277, opened 2025-08-09
wip: scroll into view on change prop

petyosi:
Related to #1276 . It's less invasive and aligns with the `followOutput` prop design.


petyosi:
@langleyd @ara4n Please review this alternative implementation  and let me know if this works for you. I've intentionally exposed the `context` parameter so that there's an additional way for the callback to determine what's to be done.

langleyd:
Ha, makes sense. Thanks for all your help!

Issue #1299, opened 2025-10-21
`GroupedVirtuoso`, `fixedItemHeight`, but for groups?


widavies:
Is there a way to set a separate `fixedItemHeight` for items and groups? Something like `fixedGroupHeight`?

My use case is identical to [this example from the docs](https://virtuoso.dev/scroll-to-group/). In this example, the "letter groups" are short, but the items themself are taller. I know this isn't trivial, willing to pay for this.

Not sure if there's an existing way to do it, I did see `itemSize`, but not sure if that's what I want here.

I believe this is the source of a mild bug:
- I supply a new `groupCounts` to `GroupedVirtuoso` with 5 additional items. Each item is 56px tall. Group headers are 28px tall. Items are added to an existing group, no new groups are added.
- `totalListHeightChanged` runs with `+280px`  (5 * 56) and then runs _again_ with `-28` px. This `-28px` is bizarre to me, oddly it's equal to the group height. I'm not sure why it runs twice here. This goes away when I set `fixedItemHeight`, hence the request above to allow for a different setting for items/groups.

<img width="764" height="137" alt="Image" src="https://github.com/user-attachments/assets/b2ca2385-1598-424a-958c-053e4a4f0279" />

<img width="294" height="45" alt="Image" src="https://github.com/user-attachments/assets/27349ef8-6e30-4520-92a7-5825501f4c28" />

I'm also willing to contribute a PR for this.

PR #1300, opened 2025-10-27
Add `fixedGroupHeight`

widavies:
@petyosi 

Here's my first go at #1299

It's a very minimal solution, but it's working for my use case. 

I have a few questions and design decisions I'd like to run past you:
1) Would you like any unit tests written to test this new prop?
2) Currently, `fixedGroupHeight` does nothing unless `fixedItemHeight` is also set. I figured there's no real reason to have `fixedGroupHeight` set but not `fixedItemHeight`
3) The specific block of code that handles fixed item sizing (`sizeSystem.ts` [line 381](https://github.com/widavies/react-virtuoso/blob/2c3698e5ab5cbe16718710a3d1aa8014e36fd31c/packages/react-virtuoso/src/sizeSystem.ts#L381)) was a little finicky. It only runs once to initialize `sizeRanges`, but I need to wait until both `fixedItemSize` and `fixedGroupSize` have been published to. To do this, I moved the `fixedGroupSize` field above the `fixedItemHeight` prop in the optional props declaration so it gets published to first & I fetch the group size with `u.getValue(..)`. Is there a better approach here?
4) Is there anywhere else this parameter needs to be handled? I canvased around a bit, it seems like `followOutputSystem.ts` may need to be updated as well?

Thanks again for taking a look.

### DIFF
--- a/.changeset/light-cobras-shine.md
+++ b/.changeset/light-cobras-shine.md
@@ -1,0 +1,5 @@
+---
+"react-virtuoso": minor
+---
+
+Support fixed size for groups in grouped virtuoso


widavies:
@petyosi It seems like there may be a bug:
- List starts empty
- Window is a certain size
- Adding items, initial fixed item size calculations are wrong


https://github.com/user-attachments/assets/3876223e-0fd1-48c2-b697-cbe676b09a85

Any ideas on this?

I'm able to temporary solve it by adding `key={totalCount === 0}` onto my `GroupedVirtuoso` so that it re-calculates on initial items getting added.

widavies:
@petyosi Alright, go ahead and take a look. Thanks!

PR #1310, opened 2025-11-30
Add `heightEstimates` prop for lists with varying item heights

### LINKED ISSUE #1308: `defaultItemHeight` as a function
In my usecase the height of the elements in the list varies widely - could be 40px, but could also be 2000px. I know up front approximately how tall each item would be - not exactly, but my approximation would be much better than taking a single item and inferring the total list height from that. The result is that currently the list scroll on the first scroll changes a lot as you scroll and discover the real height of the elements. 

If I could provide a function like `defaultItemHeight(index: number) => number` I could give it my best guess of the height of each, the initial total height calc would at least be in the same ballpark. 

I can contribute this, but wondering if you had any prior discussions/thoughts on this (couldn't find anything).

pekala:
Resolves #1308

When item heights vary widely (e.g., 40px to 2000px), the initial scrollbar size is inaccurate because virtuoso estimates total height from the first measured item. This causes noticeable scrollbar jumps during initial scrolling.

### Solution

New `heightEstimates` prop accepts an array of estimated heights per item:

```tsx
<Virtuoso
  totalCount={1000}
  heightEstimates={[40, 200, 60, 2000, 40, ...]}
  itemContent={index => <Item index={index} />}
/>
```

Estimates don't need to be exact—they seed the size tree before real measurements. Once items render, actual heights replace estimates.

### Changes

- `sizeSystem.ts`: Add `heightEstimates` stream that converts the array into size ranges (merging consecutive items with same height)
- `listSystem.ts`, `Virtuoso.tsx`: Wire up the prop
- `component-interfaces/Virtuoso.ts`: TypeScript interface + docs
- `test/sizeSystem.test.ts`: Unit tests
- `examples/height-estimates.tsx`: Interactive example with toggle to compare behavior

### DIFF
--- a/.changeset/petite-things-attack.md
+++ b/.changeset/petite-things-attack.md
@@ -1,0 +1,5 @@
+---
+"react-virtuoso": minor
+---
+
+Added heightEstimates prop to provide per-item height estimates for more accurate initial scrollbar sizing with variable height items


petyosi:
Lovely, appreciate the tests added. Thank you for the contribution, it's a very sensible one.

PR #1312, opened 2025-11-30
Add `minOverscanItemCount` prop

### LINKED ISSUE #1309: Ensure N items before/after visible viewport rendered
I'd like a prop similar to `increaseViewportBy` which ensures that there are always N items before/after the visible viewport boundaries rendered. My usecase is that I have very tall items, much taller than the viewport, and I can collapse them. When I do it, because the other items can easily fall outside of the synthetic viewport, I often see a few frames of flash of empty area, since the items below are not yet rendered. 

Is it something you'd consider adding to the project? Any prior art/discussions?

pekala:
Resolves #1309

Adds a new prop that guarantees a minimum number of items are rendered before/after the viewport, regardless of item size.

### Problem
With very tall items (taller than viewport), collapsing them causes a flash of empty space because pixel-based `increaseViewportBy` doesn't guarantee neighbouring items are pre-rendered.

### Solution
`minOverscanItemCount` ensures N items are always rendered beyond the viewport boundaries by item count rather than pixels.

```tsx
<Virtuoso minOverscanItemCount={5} />
// or per-direction:
<Virtuoso minOverscanItemCount={{ top: 2, bottom: 5 }} />
```

### Changes
- Add `minOverscanItemCount` prop to `Virtuoso` and `TableVirtuoso`
- Extend list state calculation to prepend/append overscan items
- Add documentation comparing all three overscan props
- Add example demonstrating the feature

https://github.com/user-attachments/assets/2b534dce-de63-407f-86c5-d40699c796b4

### DIFF
--- a/.changeset/bitter-glasses-arrive.md
+++ b/.changeset/bitter-glasses-arrive.md
@@ -1,6 +1,5 @@
 ---
 "react-virtuoso": minor
-"@virtuoso.dev/virtuoso.dev": patch
 ---
 
-Add minOverscanItemCount prop to ensure a minimum number of items are rendered before/after the viewport, useful for tall or collapsible items where pixel-based overscan is insufficient.
+Add `minOverscanItemCount` prop to ensure a minimum number of items are rendered before/after the viewport, useful for tall or collapsible items where pixel-based overscan is insufficient.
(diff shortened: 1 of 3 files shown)


pekala:
@petyosi Maybe we can sneak this in the new release too if you think it makes sense

petyosi:
@pekala I will need some time to review that properly, it looks great on first sight but I want to ensure I'm not missing something. 

P.S. I'm fighting NPM provenance setup at the moment, will release your other PR shortly.

PR #1444, opened 2026-06-25
fix: harden scroll-target and initial-index handling

Luccas-carvalho:
Per the request to group the related scroll-target / initial-index fixes, this PR now bundles all three. They touch the same scroll path and are described together in a single changeset.

## 1. Clamp the initial top-most item index
`getInitialTopMostItemIndexNumber` returned the raw index without bounds checking. An out-of-range `initialTopMostItemIndex` (e.g. after the data set shrinks, or a restored index that no longer fits `totalCount`) started the initial render at a blank or mid-list position. This is the initial-render counterpart of the scroll-target clamp added in #1442. Fix: clamp to `[0, lastIndex]`.

## 2. Don't mutate the location passed to scrollToIndex
`normalizeIndexLocation` applied its `align` / `behavior` / `offset` defaults directly onto the caller's location object, so a shared or memoized object was mutated (and re-published on retries). Fix: apply defaults to a shallow copy.

## 3. Treat initialTopMostItemIndex `{ index: 0 }` the same as `0`
The `didMount` filters compared the raw value against the number `0`, so the object form `{ index: 0 }` was treated as a non-default location — triggering a redundant initial scroll and delaying `followOutput`. Fix: add `initialTopMostItemIndexIsZero` and use it in both filters. `'LAST'` and non-zero indexes are unaffected.

## Tests
- `test/initialTopMostItemIndexSystem.test.ts` — clamping of `getInitialTopMostItemIndexNumber`.
- `test/scrollToIndexSystem.test.ts` — `normalizeIndexLocation` immutability and defaults.
- `test/initialTopMostItemIndexIsZero.test.ts` — numeric vs object zero, non-zero, and `'LAST'`.

Closes #1445, closes #1446.

### DIFF
--- a/.changeset/clamp-initial-topmost-index.md
+++ b/.changeset/clamp-initial-topmost-index.md
@@ -1,5 +1,0 @@
----
-"react-virtuoso": patch
----
-
-Clamp the initial top-most item index to the available range. An out-of-range `initialTopMostItemIndex` (for example after the data set shrinks) no longer starts the list at a blank or mid-list position.
--- a/.changeset/clamp-initial-topmost-index.md
+++ b/.changeset/clamp-initial-topmost-index.md
@@ -1,0 +1,5 @@
+---
+"react-virtuoso": patch
+---
+
+Clamp the initial top-most item index to the available range. An out-of-range `initialTopMostItemIndex` (for example after the data set shrinks) no longer starts the list at a blank or mid-list position.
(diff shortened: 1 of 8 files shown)


Luccas-carvalho:
@petyosi Done — folded all three into #1444. The two related fixes (the scrollToIndex location mutation and the { index: 0 } handling) now sit alongside the initial-index clamp, all described in a single changeset. Closed #1445 and #1446. Thanks for the review!

PR #1445, opened 2026-06-25
fix: avoid mutating the location passed to scrollToIndex

Luccas-carvalho:
## Problem
`normalizeIndexLocation` applied its `align` / `behavior` / `offset` defaults directly onto the caller's location object:

```ts
const result = typeof location === 'number' ? { index: location } : location
result.align = 'start' // mutates the caller's object
```

When a `scrollToIndex` consumer passes a shared or memoized location object (common in React), that object is mutated as a side effect. It also matters internally: on a scroll retry the same object is re-published, now carrying the previously-applied defaults.

## Fix
Apply the defaults to a shallow copy:

```ts
const result = typeof location === 'number' ? { index: location } : { ...location }
```

Behavior is unchanged for callers; the input object is simply left untouched.

## Tests
Added `test/scrollToIndexSystem.test.ts` asserting the input object is not mutated, that a fresh object with defaults is returned, that explicit fields are preserved, and that numeric locations are wrapped correctly.

### DIFF
--- a/.changeset/normalize-index-location-clone.md
+++ b/.changeset/normalize-index-location-clone.md
@@ -1,0 +1,5 @@
+---
+"react-virtuoso": patch
+---
+
+Stop mutating the location object passed to `scrollToIndex`. `normalizeIndexLocation` now applies its defaults to a shallow copy, so a shared or memoized location object is no longer altered (and re-published with mutated defaults on scroll retries).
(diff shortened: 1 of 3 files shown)


Luccas-carvalho:
Folding this into #1444 as requested — grouped with the related scroll-target / initial-index fixes under a single changeset.

PR #1446, opened 2026-06-25
fix: treat initialTopMostItemIndex { index: 0 } the same as 0

Luccas-carvalho:
## Problem
The `didMount` filters in `initialTopMostItemIndexSystem` compared the raw `initialTopMostItemIndex` value against the number `0`:

```ts
u.filter(([_, location]) => location !== 0)
```

When the prop is passed in object form — `initialTopMostItemIndex={{ index: 0 }}` — the comparison is `{ index: 0 } !== 0`, which is always `true`. So `{ index: 0 }` is treated as a non-default initial location: it triggers a redundant initial scroll-to-top and flips `scrolledToInitialItem` to `false`, which in turn delays `followOutput` from engaging. The numeric `0` and the object `{ index: 0 }` should behave identically.

## Fix
Add a small `initialTopMostItemIndexIsZero` helper that recognizes both the numeric and object zero, and use it in the two filters:

```ts
export function initialTopMostItemIndexIsZero(location: FlatIndexLocationWithAlign | number): boolean {
  return typeof location === 'number' ? location === 0 : location.index === 0
}
```

`{ index: 'LAST' }` and any non-zero index are unaffected and still trigger the initial scroll.

## Tests
Added `test/initialTopMostItemIndexIsZero.test.ts` covering numeric `0`, `{ index: 0 }` (with and without extra fields), non-zero numeric and object indexes, and `'LAST'`.

### DIFF
--- a/.changeset/initial-topmost-index-zero-object.md
+++ b/.changeset/initial-topmost-index-zero-object.md
@@ -1,0 +1,5 @@
+---
+"react-virtuoso": patch
+---
+
+Treat an `initialTopMostItemIndex` of `{ index: 0 }` the same as `0`. Previously the object form was considered a non-default initial location, which triggered a redundant initial scroll and delayed `followOutput` from engaging. Both forms now resolve to "start at the top".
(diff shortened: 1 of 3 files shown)


petyosi:
Can you group the last 3 PRs in one? The changes are fairly small and related, they can be well described through a changeset. Happy to close the last 2 and put everything in the first one.

Luccas-carvalho:
Folding this into #1444 as requested — grouped with the related scroll-target / initial-index fixes under a single changeset.

PR #1458, opened 2026-07-15
fix: handle dynamic horizontal direction changes

### LINKED ISSUE #1407: [BUG]: Horizontal Virtuoso does not update RTL scrollLeft normalization when CSS direction changes after mount
**Describe the bug**

I’m using Virtuoso with `horizontalDirection` for a manga/comic reader. The reader supports toggling reading direction between `LTR` and `RTL` at runtime.

When the component first mounts in `LTR` and I later toggle to `RTL`, only the initially rendered/overscanned items continue to render. The full data array is present, but scrolling toward the end does not cause more items to render.

In my real case, the list has 94 pages, but after toggling to `RTL` only about 14 pages render and the virtual range stops advancing.

**Reproduction**

[Repro](https://codesandbox.io/p/sandbox/c2hww4)

**To Reproduce**

1. Mount the list in LTR.
2. Toggle to RTL.
3. Scroll horizontally.
4. Observe that the rendered range may not advance correctly.

**Expected behavior**

When the scroll container’s CSS direction changes from ltr to rtl, horizontal Virtuoso should update its RTL scroll normalization and continue rendering items as the user scrolls.

**Screenshots**

Image

**Desktop:**

- OS: Windows.
- Browser: chrome.
- react-virtuoso: `4.17.8`

**Additional context**

- Likely Cause

`react-virtuoso` caches RTL detection in `src/utils/horizontalScroll.ts`:

```ts
const cachedRtlElements = new WeakMap<HTMLElement, boolean>()

function isRtl(scroller: HTMLElement | Window) {
  const element = scrollerElement(scroller)
  const cached = cachedRtlElements.get(element)

  if (cached !== undefined) {
    return cached
  }

  const value =
    element.ownerDocument.defaultView!.getComputedStyle(element).direction === 'rtl'

  cachedRtlElements.set(element, value)
  return value
}
```
The cache is cleared on mount/unmount, but not when the computed CSS direction changes while the same scroller element stays mounted.

- Suggested Fixes
One option is to avoid caching computed direction, since it can be inherited and changed dynamically:

```ts
function isRtl(scroller: HTMLElement | Window) {
  const element = scrollerElement(scroller)
  return element.ownerDocument.defaultView!.getComputedStyle(element).direction === 'rtl'
}
```
Another option would be to expose an imperative API to invalidate the horizontal scroll direction cache, for example:

```ts
virtuosoRef.current?.invalidateScrollDirectionCache()
```
However, I think removing or automatically invalidating the cache may be safer because CSS direction can change via props, parent dir, class names, themes, or media/state changes outside Virtuoso.

- Workaround
For now, forcing a remount when direction changes works:

```ts
<Virtuoso key={rtl ? "rtl" : "ltr"} horizontalDirection />
```
But this is not ideal because it resets internal Virtuoso state and makes direction toggling harder to preserve cleanly.

wanxiankai:
## Summary

- normalize browser-native negative RTL offsets without relying on a cached direction
- read the current computed direction when translating logical offsets for imperative scrolling
- add unit and browser regression coverage for changing a horizontal list from LTR to RTL
- include a patch changeset for react-virtuoso

Fixes #1407

## Testing

- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test` (196 passed, 4 skipped)
- `pnpm build`
- `pnpm e2e` (42 passed, 1 skipped)

### DIFF
--- a/.changeset/tidy-tools-smile.md
+++ b/.changeset/tidy-tools-smile.md
@@ -1,0 +1,5 @@
+---
+'react-virtuoso': patch
+---
+
+Handle horizontal list direction changes between LTR and RTL without remounting the list.
--- a/packages/react-virtuoso/e2e/horizontal-rtl.test.ts
+++ b/packages/react-virtuoso/e2e/horizontal-rtl.test.ts
@@ -56,17 +56,22 @@ test.describe('horizontal list with a dynamic direction', () => {
     await navigateToExample(page, baseURL, 'horizontal-dynamic-direction')
   })
 
-  test('updates logical offsets after switching from ltr to rtl', async ({ page }) => {
+  test('updates logical and physical offsets when the direction changes', async ({ page }) => {
     const scroller = page.locator('[data-testid=virtuoso-scroller]')
 
     await expect(page.locator('[data-testid=direction]')).toHaveText('ltr')
     await page.click('#toggle-direction')
     await expect(page.locator('[data-testid=direction]')).toHaveText('rtl')
 
-    await scroller.evaluate((element) => {
-      element.scrollLeft = -1000
-    })
+    await page.click('#start-20')
+    await expect(page.locator('[data-testid=range]')).toHaveText('20-25')
+    expect(await scroller.evaluate((element) => element.scrollLeft)).toBeLessThan(0)
 
+    await page.click('#toggle-direction')
+    await expect(page.locator('[data-testid=direction]')).toHaveText('ltr')
+
+    await page.click('#start-10')
     await expect(page.locator('[data-testid=range]')).toHaveText('10-15')
+    expect(await scroller.evaluate((element) => element.scrollLeft)).toBeGreaterThan(0)
   })
 })
--- a/packages/react-virtuoso/e2e/horizontal-rtl.test.ts
+++ b/packages/react-virtuoso/e2e/horizontal-rtl.test.ts
@@ -50,3 +50,23 @@ test.describe('horizontal list in rtl', () => {
     expectClosePosition(position.itemLeft, position.scrollerLeft)
   })
 })
+
+test.describe('horizontal list with a dynamic direction', () => {
+  test.beforeEach(async ({ baseURL, page }) => {
+    await navigateToExample(page, baseURL, 'horizontal-dynamic-direction')
+  })
+
+  test('updates logical offsets after switching from ltr to rtl', async ({ page }) => {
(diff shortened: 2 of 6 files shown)


petyosi:
One concern before merging: the new `Math.abs(scrollLeft)` normalization can regress Safari elastic overscroll.

Safari exposes out-of-range `scrollLeft` values during rubber-band scrolling. At the logical start, that can mean a negative value in LTR or a positive value in RTL. Previously, the direction-aware conversion kept those values negative and the downstream `Math.max(scrollTop, 0)` clamped them to zero. With `Math.abs`, they become positive logical offsets, so Virtuoso can briefly report or render a later range during bounce. MDN documents Safari's out-of-range behavior: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft

The new “normalizes negative rtl offsets” unit test also creates a default-LTR element and never sets its direction to RTL, so it currently codifies negative LTR input becoming positive.

This is the key tradeoff with avoiding the computed-style lookup on the scroll path: sign-only normalization is not semantically equivalent. I think this needs either reliable live direction information or explicit overscroll handling.

A couple of non-blocking coverage improvements:

- After changing direction, exercise `scrollToIndex` or `scrollBy` in the browser test, rather than only assigning `scrollLeft`, so the physical-write path is covered.
- Toggle back from RTL to LTR at a non-zero offset.
- There are currently no GitHub checks reported; the validations listed in the PR body are not backed by Actions results.

Apart from that, the changeset, example integration, API surface, and caller paths look reasonable.

wanxiankai:
Thanks for catching this. I pushed cff3532 to address it.

- Replaced the sign-only `Math.abs` normalization with direction-aware conversion using the current computed direction on both logical reads and physical writes. This preserves Safari elastic overscroll values as negative logical offsets, so the existing clamp still handles them.
- Updated the unit coverage to attach an actual scroller, switch its direction from LTR to RTL, and assert both normal and elastic-overscroll signs.
- Updated the browser regression to use `scrollToIndex` after LTR → RTL, assert the RTL physical write is negative, then toggle back to LTR at a non-zero offset, use `scrollToIndex` again, and assert the LTR write is positive.

Re-ran:
- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test` (196 passed, 4 skipped)
- `pnpm build`
- `pnpm e2e` (42 passed, 1 skipped)

GitHub still is not reporting checks for this PR; the results above are the local validations I re-ran after the update.

PR #1493, opened 2026-09-04
fix: apply prepend compensation in the same frame as the deviation

albertcalasanzs:
## The bug

Prepending a page blanks the transcript for a frame.

The compensation is applied in two steps one frame apart: `deviation` makes room by pushing the list
**down** (`margin-top`), and `scrollBy` cancels that push a `requestAnimationFrame` later. In between
there is a painted frame in which the list is displaced by the full size of the page with nothing
compensating it — and because pagination fires near `scrollTop: 0`, that displacement is the whole
viewport.

Reproduced frame by frame in this repo's own e2e example, in a production app, and in a design
system that bundles this package.

## The fix

The compensating scroll cannot simply move into the same task as the deviation. `deviation` reaches
the DOM through React state — `List` renders it as `marginTop`, and the synchronous style write
removed in b32e5fc is still there, commented out, directly above `setDeviation`
([Virtuoso.tsx#L107-L108](https://github.com/petyosi/react-virtuoso/blob/da1420a12bd7d103857e4828a934b0d56256298a/packages/react-virtuoso/src/Virtuoso.tsx#L107-L108)).
`scrollBy` on the other hand has a single route to the element, `scrollerRef.current.scrollBy(...)`
in `useScrollTop`. So in one task the scroll runs first, against a scroller whose content has not
grown, and the browser clamps it to the old `maxScrollTop`. On a large prepend into a short list —
"prepend 100 items before 20" — most of the compensation is silently lost, which is the upward jump
the deferral exists to prevent.

So the ordering is kept, and made structural rather than timing-based:

- `domIOSystem` gains a `deviationCommitted` stream.
- `List` and `TableVirtuoso` publish it from a `useIsomorphicLayoutEffect` keyed on the committed
  deviation — after the DOM mutation, before paint.
- `upwardScrollFixSystem` marks the compensation pending **before** publishing the deviation, then
  scrolls when an acknowledgement carrying that offset arrives. Pending-first matters: under legacy
  React outside a batch, the commit — and with it the acknowledgement — can happen synchronously
  inside the publish itself.
- If nothing acknowledges by the next frame (no renderer mounted, a subtree without layout effects,
  SSR), it falls back to the deferred scroll. The worst case is the behaviour before this change,
  never worse.

The acknowledgement lives in `List` because its `marginTop` is what grows the content in both
scroller modes, and the scroller height is additive. `List`'s deviation is a local `useState` fed by
`useEmitter`, not `useEmitterValue`, so the subscription path is identical on React 16 through 19 and
does not touch the `useSyncExternalStore` branch in `react-urx`.

## Limits, stated as limits

A pending compensation only waits for an acknowledgement when its publish actually changes the
rendered deviation. A second prepend of the same size before the first is cleared produces no new
commit, and since acknowledgements are matched by value it could be released by the previous
prepend's — so those take the fallback instead. The blank frame is therefore **not** removed in that
specific case. Removing it properly appears to require deciding whether overlapping prepends should
accumulate the deviation, which this PR does not touch.

## Tests

The previous unit test asserted that the scroll was published in the same task as the deviation,
which pinned the unsafe ordering. It now covers the protocol only — pending, released exactly once by
either path, ignoring the deviations the resize and mobile-Safari paths publish, and not releasing a
repeated prepend on the previous acknowledgement — and states that jsdom can answer neither DOM
ordering nor retained position.

The DOM-level guard is in `e2e/prepend-items.test.ts`:

- The existing retained-position assertions are **untouched**, since a clamped compensation is
  exactly what they break. Both still pass.
- A new case samples every animation frame across a prepend and fails if any painted frame had rows
  rendered while none of them intersected the viewport.

Run with `--repeat-each 20`, because the blank is a race:

| | blank frames |
|---|---|
| `main` | **20 of 20 runs** (one blank frame each) |
| this branch | **0 of 20 runs** |

Locally: `tsgo --noEmit` clean, vitest 203 passed, playwright 44 passed.

## Unrelated observation

Adding a `useWindowScroll` prepend example with a retained-position test did not work out, and the
reason may be worth knowing. On `main`, prepending 200 items grows the document from 5,206 to
15,891px but `window.scrollY` stays at 0 — a clamp would have left it at 4,486 — so the compensation
does not appear to fire at all in window-scroll mode. It behaves identically before and after this
change, so the test was dropped rather than shipped red. There is no e2e coverage for window-scroll
prepend today. Happy to open a separate issue with the repro.

### DIFF
--- a/.changeset/quiet-pandas-compensate.md
+++ b/.changeset/quiet-pandas-compensate.md
@@ -1,0 +1,16 @@
+---
+'react-virtuoso': patch
+---
+
+Fix the transcript blanking for a frame when an older page is prepended.
+
+The compensation for a prepend was a deviation that grows the content followed, one
+`requestAnimationFrame` later, by the scroll that cancels it. In between there is a painted frame in
+which the list is displaced by the whole page — and because prepends fire near `scrollTop` 0, that
+displacement is the entire viewport.
+
+The scroll now runs as soon as the renderer acknowledges, from a layout effect, that the deviation
+has reached the DOM: after the mutation, before paint. Both land in the same paint, and because the
+content has already grown the scroll can no longer be clamped to the old maximum. If nothing
+acknowledges by the next frame the previous deferred behaviour still applies, so the worst case is
+unchanged.
--- a/packages/react-virtuoso/e2e/prepend-items.test.ts
+++ b/packages/react-virtuoso/e2e/prepend-items.test.ts
@@ -34,4 +34,52 @@ test.describe('list with prependable items', () => {
 
     expect(await getScrollTop(page)).toBe(200 * 55)
   })
+
+  /**
+   * Samples every animation frame across a prepend and returns how many of them
+   * had rows rendered while none of them intersected the viewport.
+   */
+  async function blankFramesDuringPrepend(page: Page) {
+    await page.evaluate(() => {
+      const scroller = document.querySelector('[data-testid=virtuoso-scroller]')
+      if (!scroller) throw new Error('no scroller')
+      const w = window as unknown as { __blankFrames: number; __stopSampling: () => void }
+      w.__blankFrames = 0
+      let running = true
+      const sample = () => {
+        if (!running) return
+        const view = scroller.getBoundingClientRect()
+        const rows = Array.from(scroller.querySelectorAll('[data-index]'))
+        if (rows.length > 0) {
+          const anyVisible = rows.some((row) => {
+            const r = row.getBoundingClientRect()
+            return r.bottom > view.top && r.top < view.bottom
+          })
+          if (!anyVisible) w.__blankFrames += 1
+        }
+        requestAnimationFrame(sample)
+      }
+      requestAnimationFrame(sample)
+      w.__stopSampling = () => {
+        running = false
+      }
+    })
+
+    await page.locator('data-testid=prepend-200').click()
+    await page.waitForTimeout(400)
+
+    return page.evaluate(() => {
+      const w = window as unknown as { __blankFrames: number; __stopSampling: () => void }
+      w.__stopSampling()
+      return w.__blankFrames
+    })
+  }
+
+  test('never paints a frame with every row out of view', async ({ page }) => {
+    // The compensation is a deviation that grows the content plus a scroll that
+    // cancels it. If the scroll lands a frame after the growth, one painted
+    // frame has the whole list pushed below the fold — and because prepends
+    // fire near scrollTop 0, that is the entire viewport.
+    expect(await blankFramesDuringPrepend(page)).toBe(0)
+  })
 })
--- a/packages/react-virtuoso/e2e/prepend-items.test.ts
+++ b/packages/react-virtuoso/e2e/prepend-items.test.ts
@@ -42,12 +42,16 @@ test.describe('list with prependable items', () => {
   async function blankFramesDuringPrepend(page: Page) {
     await page.evaluate(() => {
       const scroller = document.querySelector('[data-testid=virtuoso-scroller]')
-      if (!scroller) throw new Error('no scroller')
+      if (!scroller) {
+        throw new Error('no scroller')
+      }
       const w = window as unknown as { __blankFrames: number; __stopSampling: () => void }
       w.__blankFrames = 0
       let running = true
       const sample = () => {
-        if (!running) return
+        if (!running) {
+          return
+        }
         const view = scroller.getBoundingClientRect()
         const rows = Array.from(scroller.querySelectorAll('[data-index]'))
         if (rows.length > 0) {
@@ -55,7 +59,9 @@ test.describe('list with prependable items', () => {
             const r = row.getBoundingClientRect()
             return r.bottom > view.top && r.top < view.bottom
           })
-          if (!anyVisible) w.__blankFrames += 1
+          if (!anyVisible) {
+            w.__blankFrames += 1
+          }
         }
         requestAnimationFrame(sample)
       }
--- a/packages/react-virtuoso/examples/prepend-stress-styles.ts
+++ b/packages/react-virtuoso/examples/prepend-stress-styles.ts
@@ -1,0 +1,138 @@
+export const stressStyles = `
+.prepend-stress {
+  color: #17212e;
+  font:
+    14px/1.45 system-ui,
+    sans-serif;
+}
+.prepend-stress > p {
+  max-width: 950px;
+}
+.prepend-controls {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  gap: 8px;
+}
+.prepend-controls button {
+  background: #17212e;
+  color: white;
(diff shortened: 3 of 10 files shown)


petyosi:
Thanks for investigating this. The blank frame is a real problem, but I don't think we can safely merge this ordering change as written.

I checked the history. The first `requestAnimationFrame` was introduced deliberately in b32e5fc669aa0b8c588e2b05d1dd3aa030f688e8 ("fix: prepend 100 items before 20"), immediately after the React 18 prepend work in 7ed6bc82727a4180702342c76ccff1814b625045.

That commit made two coordinated changes:

- it removed the synchronous `ref.current.style.marginTop` update and left `deviation` to be applied through React state;
- it deferred `scrollBy` by one frame.

The resulting sequence is intentional:

1. publish the deviation and let React commit the margin;
2. scroll by the same amount;
3. clear the deviation and release `recalcInProgress`, allowing the recalculated list state to render.

Publishing `scrollBy` in the same task removes the guarantee that the deviation has reached the DOM first. Depending on React scheduling, the scroll can therefore happen against the old layout. That can reintroduce the upward jump or an incorrect intermediate scroll state that the 2022 changes addressed. The package still supports React 16–19, and no later migration appears to have removed this constraint.

The new unit test verifies stream publication order, but it does not verify the required DOM order or the retained visual position. It effectively pins the potentially unsafe behavior.

Could you revise the fix so that the deviation is known to be committed before `scrollBy`, while still ensuring that both DOM effects happen before the same paint? A layout-effect acknowledgement from the list renderer, or another explicitly synchronous handoff, may be a suitable direction. The regression test should exercise the rendered DOM and verify the retained position, ideally across the supported React paths and the relevant scroller modes.

I don't want to dismiss the reported issue or transfer its investigation back to consumers. The goal is to close the blank-frame window without removing the synchronization that protects prepend positioning.

albertcalasanzs:
**TL;DR** — Publishing `scrollBy` in the same task scrolls before the margin reaches the DOM, and my A/B could not have caught it. The revision waits for a layout-effect acknowledgement from the renderer, with the previous rAF deferral as fallback. The retained-position e2e is untouched and green; a new frame sampler fails on `main` and passes here.

**Mechanism.** In the list path `deviation` reaches the DOM only through React state: `List` renders it as `marginTop`, and the synchronous write removed in b32e5fc is still there, commented out, directly above `setDeviation` ([Virtuoso.tsx#L107-L108](https://github.com/petyosi/react-virtuoso/blob/da1420a12bd7d103857e4828a934b0d56256298a/packages/react-virtuoso/src/Virtuoso.tsx#L107-L108)). `scrollBy` has a single route to the element — `scrollerRef.current.scrollBy(location)` in `useScrollTop`. So publishing both in one task is not a scheduling race: the deviation update is batched, the scroll runs first against a scroller whose content has not grown, and the browser clamps to the old maxScrollTop. On "prepend 100 items before 20" most of the compensation is lost — the upward jump the deferral prevents.

My A/B never surfaced it: it ran on an 851-message conversation at scrollTop ~29,604 of ~40,256px, so the scroll always had room. The blank-frame result stands; the safety conclusion I drew from it was outside what that test could show.

**Revision.** The ordering stays; it is now structural rather than timing-based.

- `domIOSystem` gains a `deviationCommitted` stream.
- `List` and `TableVirtuoso` publish it from a `useIsomorphicLayoutEffect` keyed on the committed deviation — after the DOM mutation, before paint.
- `upwardScrollFixSystem` marks the compensation pending *before* publishing the deviation, then scrolls when an acknowledgement carrying that offset arrives. Pending-first matters: under legacy React outside a batch the commit, and with it the ack, can happen synchronously inside the publish.
- If nothing acknowledges by the next frame (no renderer mounted, a subtree without layout effects, SSR), it falls back to the deferred scroll. Worst case is the behaviour before this change, never worse.

**One limit, stated as a limit.** A pending compensation only waits for an acknowledgement when its publish actually changes the rendered deviation. A second prepend of the same size before the first is cleared produces no new commit, and since acknowledgements are matched by value it could be released by the previous prepend's — so those take the fallback, and the blank frame is not removed in that specific case. Removing it properly seems to require deciding whether overlapping prepends should accumulate the deviation, which I have not touched.

**Tests.** You're right that the old unit test pinned the unsafe ordering. It now covers the protocol only — pending, released exactly once by either path, ignoring the deviations the resize and mobile-Safari paths publish, and not releasing a repeated prepend on the previous acknowledgement — and states that jsdom can answer neither DOM ordering nor retained position.

The retained-position assertions in `e2e/prepend-items.test.ts` are untouched, since a clamp is exactly what they break. Added to them: a frame sampler that fails if any painted frame had rows rendered while none intersected the viewport. With `--repeat-each 20` it fails **20 of 20** runs on `main` and **0 of 20** here — deterministic in that example rather than intermittent. Locally: typecheck clean, vitest 203 passed, playwright 44 passed.

Everything else — the React-version reasoning, the `TableVirtuoso` detail, and a window-scroll observation unrelated to this change — is in the PR description rather than repeated here.

petyosi:
Merged, including the stable scroll callback fix and the mixed-content Ladle stress stories. Local validation passed: package lint and typechecking, 203 unit tests, and 44 browser tests (with four unit tests and one browser test skipped).

One limitation is worth recording for future reverse-scroll work: this fixes the prepend compensation handoff, but it cannot eliminate every adjustment caused by newly measured items in Virtuoso's current layout architecture.

Virtuoso lays out its ordinary list items in normal document flow. When scrolling upward mounts an item whose actual height differs from its estimate, the browser can move the following items before the measurement has been incorporated and the scroll position compensated. In our mixed-height stress case, scrolling upward after prepending 100 items exposed an approximately 85px transient adjustment when scrolling stopped. Enabling `skipAnimationFrameInResizeObserver` did not eliminate it. An experimental synchronous layout-effect measurement reduced the sampled movement, but that would add synchronous layout work and needs separate performance and correctness work.

These are limitations of the current measurement and layout sequence, rather than something this PR's prepend fix can fully solve. Eliminating them would require stronger before-paint measurement/compensation coordination or a different positioning strategy.

**For chat interfaces, Virtuoso Message List is the recommended component.** Its positioning model is better suited to loading older messages and handling mixed-height content, and it delivered more stable reverse scrolling in this comparison than the general-purpose Virtuoso list. This is a recommendation based on chat scrolling behavior, not a claim that it wins every performance benchmark.

We compared the same content with Virtuoso Message List. Its absolutely positioned items do not automatically push their neighbors when a newly mounted item gets measured; their positions remain controlled by the layout model. That avoids the normal-flow displacement mechanism, and the same upward-scroll/stop test showed only subpixel movement. The comparison also exposed a separate fractional-boundary prepend issue in Message List, now fixed in 1.17.3.

Absolute positioning is not a guarantee against every visual artifact: Chrome can still expose a temporary unrendered area during fast scrolling, which a larger `increaseViewportBy` buffer can reduce. That is distinct from item displacement. Overlapping prepend operations also remain a separate concern.

